PR #47, opened 2013-04-18
Script-nonce support

coderanger:
An example use would be:

``` ruby
:script_nonce => lambda { params[:script_nonce] = SecureRandom.hex(16) },
```

and then in your Haml (if you are so inclined):

``` haml
%script{type: "text/javascript", nonce: params[:script_nonce]}
```

Minor downside, as far as I know no browser yet supports this for actual enforcement though WebKit itself added support [last year](https://bugs.webkit.org/show_bug.cgi?id=89577). The spec is designed for transparent upgrade though, so you can add it now and benefit silently once it is supported.

### DIFF
--- a/lib/secure_headers.rb
+++ b/lib/secure_headers.rb
@@ -72,10 +72,10 @@ module SecureHeaders
       options = self.class.options_for :csp, options
       return if options == false
 
-      header = ContentSecurityPolicy.new(options, :request => request)
+      header = ContentSecurityPolicy.new(options, :request => request, :controller => self)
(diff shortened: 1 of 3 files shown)


oreoshake:
@coderanger if you have any interest, I suggest you apply for a Google Patch Award. Adding nonce support to CSP helpers is explicitly listed as an acceptable change: https://www.google.com/about/appsecurity/patch-rewards/

coderanger:
@oreoshake Thanks for the heads up!

Issue #130, opened 2015-01-15
Rails 4 support


naiyt:
It seems like secureheaders has some issues on Rails 4 (tested on Rails 4.1.6). The headers being set in my initializer aren't what's actually being served. From what I've read, Rails 4 sets more default headers now, like X-Frame-Options. My assumption is that secureheaders is setting X-Frame-Options first, but then Rails is overwriting it at a later point.

Is this a known issue? If other people are having this issue (and it's not just something to do with how I have it configured) I could look into it and try to create a PR.

oreoshake:
Yeah I had originally proposed that in twitter/secureheaders/issues/69 but never got around to it. I'd love a pull request if you could :smile_cat:

naiyt:
Sure, I'll take a look and see if I can figure out it!

oreoshake:
I think it's just a call to `config.action_dispatch.default_headers.clear` in an initializer block

stve:
> I think it's just a call to `config.action_dispatch.default_headers.clear` in an initializer block

That would clear out any non-security related headers as well as an unintended side-effect, presumably we'd need to only remove headers that are defined in this gem.

oreoshake:
> we'd need to only remove headers that are defined in this gem

Good call.

naiyt:
Yeah, good point., @stve. I'm working on getting a failing spec written right now.

naiyt:
I added a [Rails 4 fixture](https://github.com/naiyt/secureheaders/compare/rails4) trying to match as closely as possible what you have for the Rails 3 fixtures. Oddly, though, I can't seem to get replicate the issue when running those specs.

To test, I set `config.x_frame_options = 'DENY'` [in the initializer](https://github.com/naiyt/secureheaders/blob/550ca519c458fdf5bc041aab67cde80dc86781e0/fixtures/rails_4_1_8/config/initializers/secure_headers.rb#L3). ActionDispatch defaults it to `SameOrigin`.

So I would expect [that this spec should fail](https://github.com/naiyt/secureheaders/blob/550ca519c458fdf5bc041aab67cde80dc86781e0/fixtures/rails_4_1_8/spec/controllers/other_things_controller_spec.rb) on Rails 4. However, it passes, which means secureheaders is setting that header correctly.

Either some quirk of the test environment, or I'm doing something silly. I'll keep looking into it, but if you guys notice anything obvious I'm missing let me know. :+1:

naiyt:
[...]

PR #131, opened 2015-01-16
set the header in default_headers if defined there

naiyt:
@oreoshake how do you feel about this as a fix for #130 ? Seems to work fine for me when I test it manually, and only overwrites a header in `config.action_dispatch.default_headers` if we're explicitly setting it with `secureheaders`.

If it looks like an okay solution I'll add in specs.


naiyt:
Hm, something else I just thought of is the `skip_before_filters`. Doing, for example `skip_before_filter :set_x_frame_options_header` will cause it to return to the Rails default of `SAMEORIGIN` as opposed to removing the header, which would probably be what you expect.

oreoshake:
Good point. We could clear just the headers that this gem sets, and then things would be have as expected.

lib/secure_headers/railtie.rb
```diff
@@ -3,9 +3,17 @@
   module SecureHeaders
     class Railtie < Rails::Engine
       isolate_namespace ::SecureHeaders if defined? isolate_namespace # rails 3.0
+      conflicting_headers = ['X-Frame-Options', 'X-XSS-Protection', 'X-Content-Type-Options']
       initializer "secure_headers.action_controller" do
         ActiveSupport.on_load(:action_controller) do
           include ::SecureHeaders
+
+          unless Rails.application.config.action_dispatch.default_headers.nil?
```
naiyt: This seems to work correctly, and allows stuff like the `skip_before_filters` to still work as expected.

naiyt:
Good idea, @oreoshake. I've done that in the initializer and it seems to work fine. I've also added another fixture for Rails 4 matching what you had for the others, but setting the headers that Rails sets to something else to ensure this fix is working.

If this looks good, let me know if you'd prefer me to squash the commit history.

lib/secure_headers/railtie.rb
```diff
@@ -3,9 +3,17 @@
   module SecureHeaders
     class Railtie < Rails::Engine
       isolate_namespace ::SecureHeaders if defined? isolate_namespace # rails 3.0
+      conflicting_headers = ['X-Frame-Options', 'X-XSS-Protection', 'X-Content-Type-Options']
```
oreoshake: I think we should future proof this by including the headers this gem sets, even if they aren't in default headers.

i.e. csp, csp report only, hpkp, xdownload, etc

lib/secure_headers/railtie.rb
```diff
@@ -3,9 +3,22 @@
   module SecureHeaders
     class Railtie < Rails::Engine
       isolate_namespace ::SecureHeaders if defined? isolate_namespace # rails 3.0
+      conflicting_headers = ['X-Frame-Options', 'X-XSS-Protection', 'X-Content-Type-Options',
+                             'X-Permitted-Cross-Domain-Policies', 'X-Download-Options',
+                             'X-Content-Type-Options', 'Strict-Transport-Security',
+                             'Content-Security-Policy', 'Content-Security-Policy-Report-Only',
+                             'X-Permitted-Cross-Domain-Policies']
```
naiyt: Good point on that @oreoshake RE: adding the other headers to future proof it. Might also be nice if there was a way to reach in and grab the header names directly from each `Header` class, but I couldn't think of a good way to do it.

oreoshake: Yeah, each header class does have a name constant, but with CSP it's a little awkward. I think the string values are fine.

Please add the hpkp headers too twitter/secureheaders/pull/132

naiyt:
HPKP headers are in now.

oreoshake:
Hmm that test failure might be annoying. Rails 4 doesn't support 1.8.7 and it generates a mixed bag of hash syntaxes. Would you mind converting those? I can take a stab at it, but probably won't get to it.

naiyt:
No problem, I'll do that later today.

naiyt:
Luckily the only thing in the Rails 4 fixture not using the hash rocket was the route, so it passes on 1.8.7 now.

lib/secure_headers/railtie.rb
oreoshake: I feel like this would raise an error on rails < 4, default_headers wouldn't even be defined! Must be more rails method missing crap magic.

oreoshake:
:sparkles:

oreoshake:
I'll release a gem tomorrow

PR #132, opened 2015-01-17
Add HPKP support

thirstscolr:
I've added support for HPKP headers now that it is supported in both Chrome and Firefox.

### DIFF
--- a/README.md
+++ b/README.md
@@ -8,6 +8,7 @@ The gem will automatically apply several headers that are related to security.  
 - X-Content-Type-Options - [Prevent content type sniffing](http://msdn.microsoft.com/en-us/library/ie/gg622941\(v=vs.85\).aspx)
 - X-Download-Options - [Prevent file downloads opening](http://msdn.microsoft.com/en-us/library/ie/jj542450(v=vs.85).aspx)
 - X-Permitted-Cross-Domain-Policies - [Restrict Adobe Flash Player's access to data](https://www.adobe.com/devnet/adobe-media-server/articles/cross-domain-xml-for-streaming.html)
+- Public Key Pinning - Pin certificate fingerprints in the browser to prevent man-in-the-middle attacks due to compromised Certificate Authorites. [Public Key Pinnning  Specification](https://tools.ietf.org/html/draft-ietf-websec-key-pinning-21)
 
 ## Usage
 
@@ -21,6 +22,7 @@ The following methods are going to be called, unless they are provided in a `ski
 
 * `:set_csp_header`
 * `:set_hsts_header`
+* `:set_hpkp_header`
 * `:set_x_frame_options_header`
 * `:set_x_xss_protection_header`
 * `:set_x_content_type_options_header`
@@ -50,6 +52,12 @@ This gem makes a few assumptions about how you will use some features.  For exam
     :frame_src => "https: http:.twimg.com http://itunes.apple.com",
     :img_src => "https:",
     :report_uri => '//example.com/uri-directive'
+  }
+  config.hpkp = {
+    :max_age => 60.days.to_i,
+    :include_subdomains => true,
+    :report_uri => '//example.com/uri-directive',
+    :pins => [{:sha256 => 'b5bb9d8014a0f9b1d61e21e796d78dccdf1352f23cd32812f4850b878ae4944c'}]
   }
 end
 
@@ -298,6 +306,24 @@ console.log("will raise an exception if not in script_hashes.yml!")
 <% end %>
 ```
 
+### Public Key Pins
+
+```
+config.hpkp = {
+  max_age: 60.days.to_i,   # max_age is a required parameter
+  include_subdomains: true, # whether or not to apply pins to subdomains
+  # Per the spec, SHA256 hashes are the only currently supported format.
+  pins: [
+    {sha256: 'b5bb9d8014a0f9b1d61e21e796d78dccdf1352f23cd32812f4850b878ae4944c'},
+    {sha256: '73a2c64f9545172c1195efb6616ca5f7afd1df6f245407cafb90de3998a1c97f'}
+  ],
+  enforce: true,            # defaults to false (report-only mode)
+  report_uri: '//example.com/uri-directive',
+  app_name: 'example',
+  tag_report_uri: true
+}
+```
+
 ### Using with Sinatra
 
 Here's an example using SecureHeaders for Sinatra applications:
@@ -321,6 +347,7 @@ require 'secure_headers'
     :img_src => "https: data:",
     :frame_src => "https: http:.twimg.com http://itunes.apple.com"
   }
+  config.hpkp = false
 end
 
 class Donkey < Sinatra::Application
--- a/lib/secure_headers.rb
+++ b/lib/secure_headers.rb
@@ -6,7 +6,7 @@ module SecureHeaders
     class << self
       attr_accessor :hsts, :x_frame_options, :x_content_type_options,
         :x_xss_protection, :csp, :x_download_options, :script_hashes,
-        :x_permitted_cross_domain_policies
+        :x_permitted_cross_domain_policies, :hpkp
 
       def configure &block
         instance_eval &block
@@ -42,6 +42,7 @@ module SecureHeaders
       self.secure_headers_options = options
       before_filter :prep_script_hash
       before_filter :set_hsts_header
+      before_filter :set_hpkp_header
       before_filter :set_x_frame_options_header
       before_filter :set_csp_header
       before_filter :set_x_xss_protection_header
@@ -61,6 +62,7 @@ module SecureHeaders
     def set_security_headers(options = self.class.secure_headers_options)
       set_csp_header(request, options[:csp])
       set_hsts_header(options[:hsts])
+      set_hpkp_header(options[:hpkp])
       set_x_frame_options_header(options[:x_frame_options])
       set_x_xss_protection_header(options[:x_xss_protection])
       set_x_content_type_options_header(options[:x_content_type_options])
@@ -136,6 +138,16 @@ module SecureHeaders
       set_a_header(:hsts, StrictTransportSecurity, options)
     end
 
+    def set_hpkp_header(options=self.class.secure_headers_options[:hpkp])
+      return unless request.ssl?
+      config = self.class.options_for :hpkp, options
+
+      return if config == false || config.nil?
+
+      hpkp_header = PublicKeyPins.new(config)
+      set_header(hpkp_header)
+    end
+
     def set_x_download_options_header(options=self.class.secure_headers_options[:x_download_options])
       set_a_header(:x_download_options, XDownloadOptions, options)
     end
@@ -168,6 +180,7 @@ end
 
 require "secure_headers/version"
 require "secure_headers/header"
+require "secure_headers/headers/public_key_pins"
 require "secure_headers/headers/content_security_policy"
 require "secure_headers/headers/x_frame_options"
 require "secure_headers/headers/strict_transport_security"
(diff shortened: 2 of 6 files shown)


lib/secure_headers/headers/public_key_pins.rb
```diff
@@ -0,0 +1,97 @@
...
+      base
+    end
+
+    def value
+      header_value = [
+        generic_directives,
+        pin_directives,
+        report_uri_directive,
+        subdomain_directive
+      ].join.strip
```
oreoshake: nit: can you `join(';')` instead of prefixing some with "; ..." ?

oreoshake:
Damn you 1.8.7. I'm ready to drop support and the timing is right given the pending 2.0 release.

lib/secure_headers.rb
```diff
@@ -136,6 +138,16 @@ def set_hsts_header(options=self.class.secure_headers_options[:hsts])
       set_a_header(:hsts, StrictTransportSecurity, options)
     end
 
+    def set_hpkp_header(options=self.class.secure_headers_options[:hpkp])
+      return unless request.ssl?
+      config = self.class.options_for :hpkp, options
+
+      return if config == false || config.nil?
```
oreoshake: nice. can you add a test ensuring this header isn't set by default?

oreoshake:
I'd love an additional sanity check from someone with experience working with the header. @ptoomey3  ? Did the pin-* value need to be quoted or something?

lib/secure_headers/headers/public_key_pins.rb
```diff
@@ -0,0 +1,97 @@
...
+      end
+
+      config
+    end
+
+    def pin_directives
+      return '' if @pins.nil?
+      @pins.collect do |pin|
+        pin.map do |token, hash|
+          "; pin-#{token}=\"#{hash}\"" if HASH_ALGORITHMS.include?(token)
```
oreoshake: please add a test for the pin values

spec/lib/secure_headers_spec.rb
```diff
@@ -196,6 +198,38 @@ def set_security_headers(subject)
     end
   end
 
+  describe "#set_public_key_pins" do
+    it "sets the Public-Key-Pins header" do
+      should_assign_header(HPKP_HEADER_NAME + "-Report-Only", "max-age=1234")
```
oreoshake: is it valid to set this without at least one pin- entry?

oreoshake:
If you merge master, the build will pass on 1.8.7

spec/lib/secure_headers_spec.rb
thirstscolr: yes -- only `max_age` is required. from http://tools.ietf.org/html/draft-ietf-websec-key-pinning-21#section-2.1.1

"""According to the processing rules of Section 2.1, the UA MUST ignore
   pin-directives with tokens naming hash algorithms it does not
   recognize.  If the set of remaining effective pin-directives is
   empty, and if the host is a Known Pinned Host, the UA MUST cease to
   consider the host as a Known Pinned Host (the UA should fail open)."""

oreoshake: :cool: , thanks for the pointer.

lib/secure_headers/headers/public_key_pins.rb
```diff
@@ -0,0 +1,97 @@
...
+    end
+
+    def build_directive(key)
+      "#{self.class.symbol_to_hyphen_case(key)}=#{@config[key]}"
+    end
+
+    def report_uri_directive
+      return '' if @report_uri.nil?
+
+      if @report_uri.start_with?('//')
```
ptoomey3: This block seems unneeded. Supporting a scheme relative URL in the config implies the resulting final URL will go over `http` or `https` based on some sort of context. But, it will always get set to go over `https`. I vote that `report_uri` is documented simply as `:report_uri => "http[s]://example.com/uri-directive"`.

lib/secure_headers/headers/public_key_pins.rb
ptoomey3: @oreoshake - This (`\"#{hash}\"`) is the quoting that is necessary in Firefox 35 and will soon be mandatory in Chrome Canary. :+1:

lib/secure_headers/headers/public_key_pins.rb
```diff
@@ -0,0 +1,97 @@
...
+        raise PublicKeyPinsBuildError.new("max-age must be a number.
+                                          #{config[:max_age]} was supplied.")
+      end
+
+      config
+    end
+
+    def pin_directives
+      return '' if @pins.nil?
+      @pins.collect do |pin|
```
ptoomey3: As noted in http://tools.ietf.org/html/draft-ietf-websec-key-pinning-21#section-4.3 and http://tools.ietf.org/html/draft-ietf-websec-key-pinning-21#section-2.5:

> The given set of Pins contains at least one Pin that does NOT
      refer to an SPKI in the certificate chain.  (That is, the host
      must set a Backup Pin;
>...
> Because having a backup key pair is so important to recovery, UAs
   MUST require that hosts set a Backup Pin 

So, `validate_config` should check to make sure that >= 2 unique pins are provided in the config.

README.md
```diff
@@ -298,6 +306,24 @@ console.log("will raise an exception if not in script_hashes.yml!")
...
+config.hpkp = {
+  max_age: 60.days.to_i,   # max_age is a required parameter
+  include_subdomains: true, # whether or not to apply pins to subdomains
+  # Per the spec, SHA256 hashes are the only currently supported format.
+  pins: [
+    {sha256: 'b5bb9d8014a0f9b1d61e21e796d78dccdf1352f23cd32812f4850b878ae4944c'},
+    {sha256: '73a2c64f9545172c1195efb6616ca5f7afd1df6f245407cafb90de3998a1c97f'}
+  ],
+  enforce: true,            # defaults to false (report-only mode)
+  report_uri: '//example.com/uri-directive',
```
ptoomey3: You may want to add a note and/or a warning that `report_uri` and `Public-Key-Pins-Report-Only` have not yet been implemented yet in Chrome: https://code.google.com/p/chromium/issues/detail?id=445793#c2 (I'm not sure of the status in Firefox). Similarly, it may be good to document/warn that setting a `report_uri` to the same domain as the hosting application can lead to unexpected behavior because, as noted in http://tools.ietf.org/html/draft-ietf-websec-key-pinning-21#section-2.1.4:

> If the scheme in
   the report-uri is one that uses TLS (e.g.  HTTPS), UAs MUST perform
   Pinning Validation when the host in the report-uri is a Known Pinned
   Host; similarly, UAs MUST apply HSTS if the host in the report-uri is
   a Known HSTS Host.

So, if you have a pin failure you will not be able to send a report back to the hosting application via `https`. Moreover, if the site is HSTS you won't be able to send back a report of `https` or `http` (because of HSTS).

ptoomey3:
I just gave that a quick glance, but left some initial feedback. More generally, I would be hesitant to set this for Chrome, as the current stable implementation has a number of bugs:

https://code.google.com/p/chromium/issues/detail?id=444511
https://code.google.com/p/chromium/issues/detail?id=412866
.. I'd have to double check if there are any additional known issues

The above bugs prevent `max-age` from working, don't let you revoke/rotate out a known pin, and always sets the `includeSubdomains` directive even when the policy does NOT have it set. So, I'd be a bit cautious with putting the directive in a release until things settle down. Supposedly these will be fixed in the next Canary release, so I haven't had a chance to validate the fixes yet. I haven't played around with Firefox 35 as much, but I haven't yet seen anything unexpected/that violates the spec.

README.md
```diff
@@ -8,6 +8,7 @@ The gem will automatically apply several headers that are related to security.
 - X-Content-Type-Options - [Prevent content type sniffing](http://msdn.microsoft.com/en-us/library/ie/gg622941\(v=vs.85\).aspx)
 - X-Download-Options - [Prevent file downloads opening](http://msdn.microsoft.com/en-us/library/ie/jj542450(v=vs.85).aspx)
 - X-Permitted-Cross-Domain-Policies - [Restrict Adobe Flash Player's access to data](https://www.adobe.com/devnet/adobe-media-server/articles/cross-domain-xml-for-streaming.html)
+- Public Key Pinning - Pin certificate fingerprints in the browser to prevent man-in-the-middle attacks due to compromised Certificate Authorites. [Public Key Pinnning  Specification](https://tools.ietf.org/html/draft-ietf-websec-key-pinning-20)
```
ptoomey3: I think draft 21 is the latest.

README.md
```diff
@@ -298,6 +306,24 @@ console.log("will raise an exception if not in script_hashes.yml!")
...
 ```
 
+### Public Key Pins
+
+```
+config.hpkp = {
+  max_age: 60.days.to_i,   # max_age is a required parameter
+  include_subdomains: true, # whether or not to apply pins to subdomains
+  # Per the spec, SHA256 hashes are the only currently supported format.
+  pins: [
```
ptoomey3: This syntax feels a little awkward (could just be me). Given that it would be extremely rare for each pin to use their own hash type, it may make sense to just set it globally:

```
...
pin_hash: "sha256",
pins: ["pin1base64hash", "pin2base64hash"],
...
```

Or, just bake the hash type into the pin

```
pins: ["sha256:pin1base64hash", "sha256:pin2base64hash"],
```

I tend to prefer the first style, as there is no reason you can't/shouldn't use the same hash format for all pins.

lib/secure_headers/headers/public_key_pins.rb
thirstscolr: :+1:

README.md
thirstscolr: SHA256 is the currently only supported hash algorithm but I wanted this to be flexible in the case it is extended.

lib/secure_headers/headers/public_key_pins.rb
thirstscolr: good catch ill add that check.

README.md
ptoomey3: Yeah, I definitely think it should be flexible enough to support alternate hash algos. I just thought the array of hashes syntax was a bit awkward. I'll leave up to @oreoshake since it is purely stylistic preference.

oreoshake: LOL `secure_headers` thrives on awkward abuse of the `Hash` class. 

But seriously, I like the "array of hashes" syntax more than either suggestion. I never said I was good at API design.

ptoomey3: Ha...fair enough :smile:

spec/lib/secure_headers_spec.rb
```diff
@@ -196,6 +211,38 @@ def set_security_headers(subject)
...
+    end
+
+    it "allows you to specify a report-uri" do
+      should_assign_header(HPKP_HEADER_NAME, "max-age=1234; report-uri=\"https://foobar.com\"")
+      subject.set_hpkp_header(max_age: 1234, report_uri: "https://foobar.com", enforce: true)
+    end
+
+    it "allows you to specify a report-uri with app_name" do
+      should_assign_header(HPKP_HEADER_NAME, "max-age=1234; report-uri=\"https://foobar.com?enforce=true&app_name=my_app\"")
+      subject.set_hpkp_header(max_age: 1234, report_uri: "https://foobar.com", app_name: "my_app", tag_report_uri: true, enforce: true)
```
oreoshake: please use the 1.8.7 hash syntax => (sorry, still need to support 1.8.7)

oreoshake:
@ptoomey3 what say you, is hpkp ready for inclusion? Any browser quirks worth working around?

ptoomey3:
I haven't actually tested it in Chrome V42 stable, but when I tested Chrome V42 beta the HPKP bug was fixed. So, in all likelihood, yeah, this should be stable enough in Chrome/Firefox to be useful.

ptoomey3:
> Any browser quirks worth working around?

I don't believe so.

README.md
oreoshake: How about something like 

> Be aware that pinning error reporting is governed by the same rules as everything else. If you have a pinning failure that tries to report back to the same host, by definition this will not work. 

@ptoomey3 Do people typically supply a separate host? Do reports go cross-origin? I forget where CSP implementations ended up, but I've had to use a proxy host in the past :sob: to get around this

PR #133, opened 2015-01-17
Make yet another exception for 1.8.7

oreoshake:
Should probably just :fire: 1.8.7 support but this works.


oreoshake:
Hmm I'm no bundler expert but it looks like this is restricting the `i18n` version for all platforms.

presidentbeef:
I think it should be `:platforms`?

oreoshake:
lulz. thanks!

PR #137, opened 2015-03-20
Include ActionView helpers in ActionView::Base

jstorimer:
Including `SecureHeaders::ViewHelpers` directly into `ActionView::Helpers` does *not* make them available in views in Rails 4.2. Including the helper module directly into `ActionView::Base` makes them available.

I'm not sure if this is going to fly in Rails 3.x, will have to wait for the results from Travis.

### DIFF
--- a/lib/secure_headers/view_helper.rb
+++ b/lib/secure_headers/view_helper.rb
@@ -62,7 +62,7 @@ module SecureHeaders
 end
 
 module ActionView #:nodoc:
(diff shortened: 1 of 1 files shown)


jstorimer:
CI seems happy with this :+1:

oreoshake:
Thanks, just released a new gem 2.0.1. Confirmed it works in rails 3 land.

PR #140, opened 2015-04-30
Update README.md

jmera:
Fix broken link.


oreoshake:
If you don't mind, use `http://www.w3.org/TR/CSP2/` and update the 1.1 reference to "level 2"

jmera:
@oreoshake made those changes and squashed.

Issue #146, opened 2015-06-17
Per instance hook


rupurt:
Howdy,

Great gem, thanks! I was wondering if there's any way to modify the headers on a per instance basis? For example using the CSP preferences for the current user.

I had a quick peek into the source code and couldn't find a hook point that would achieve this. However with a small modification I think this could be achieved. I was thinking that instead of calling [`self.class.options_for`](https://github.com/twitter/secureheaders/blob/master/lib/secure_headers.rb#L83) it could call an instance method `self.options_for` which would do the same thing. That would then allow you to use `super` and you can modify the returned options hash. Something along these lines.

```ruby
# lib/secure_headers.rb
module SecureHeaders
   module InstanceMethods
      # ... stuff

      def set_csp_header(req = nil, config=nil)
         # ... more stuff

         config = self.class.secure_headers_options[:csp] if config.nil?
         config = options_for :csp, config

         return if config == false

         # ... even more stuff
      end

      # This can now be overridden
      def options_for(type, options)
         options.nil? ? ::SecureHeaders::Configuration.send(type) : options
      end
   end
end
```

oreoshake:
That's a great idea! It's been [asked for](https://github.com/twitter/secureheaders/issues/134) before. 

BTW all csp config values accept procs which can work for simple cases. I do like this approach, but `options_for` isn't descriptive enough to plop in someone else's namespace.

This could also be extended to the rest of the config so maybe something like `def secure_headers_options` which returns the entire hash, likely a modified version of `super`'s.

wdyt?

rupurt:
Cool.

I just realized that what I'm doing doesn't need something on a per action basis. It needs it on a per instance basis. I've updated the title of the issue to reflect this. Sorry.

I agree that `options_for` is probably too generic. How do you feel about `secure_header_options_for`? I don't fully understand what you mean by `This could also be extended to the rest of the config`, do you mind elaborating? I've created a [branch](https://github.com/SchoolKeep/secureheaders/commit/1a9d2ea71a69a62aa1aa5b1694fe81fb29454d77) which reflects my thoughts.

The 2 key things this helps with are:

1. Overriding the `options_for` (or equivalent) method can use the default config and we can override specific keys instead of using `ensure_security_headers` which clobbers the default config.
2. Overriding `options_for` in the controller allows us to use per instance data e.g. `current_user`

oreoshake:
[...]

PR #147, opened 2015-06-17
Change `options_for` from class method to instance method

rupurt:
Based on the thread from #146

This will help with:

- Allowing a controller to use instance data e.g. `current_user`
- Allowing a controller instance to override the default keys set in `::SecureHeaders::Configuration.configure`

### DIFF
--- a/lib/secure_headers.rb
+++ b/lib/secure_headers.rb
@@ -50,12 +50,6 @@ module SecureHeaders
       before_filter :set_x_download_options_header
       before_filter :set_x_permitted_cross_domain_policies_header
     end
-
-    # we can't use ||= because I'm overloading false => disable, nil => default
-    # both of which trigger the conditional assignment
-    def options_for(type, options)
-      options.nil? ? ::SecureHeaders::Configuration.send(type) : options
-    end
   end
 
   module InstanceMethods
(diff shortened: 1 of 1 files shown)


oreoshake:
Can you add a test? e.g. https://github.com/twitter/secureheaders/blob/master/fixtures/rails_4_1_8/spec/controllers/things_controller_spec.rb

rupurt:
Because I moved the method from public to private I was thinking the existing test coverage would suffice. If that doesn't work what kind of a test would you like to see?

oreoshake:
In one of the fixture apps, override the default behavior for a single action that would alter the policy and verify the value on the way out. It will at least show that per-action results are possible.

e.g. add an action to https://github.com/twitter/secureheaders/blob/master/fixtures/rails_4_1_8/app/controllers/things_controller.rb, override options_for or whatever with something like `return something if params[:action] == :my_new_action`, check the policy in https://github.com/twitter/secureheaders/blob/master/fixtures/rails_4_1_8/spec/controllers/things_controller_spec.rb

oreoshake:
btw, I submitted an alternative: https://github.com/twitter/secureheaders/pull/148

I do plan on shipping both options. I like https://github.com/twitter/secureheaders/pull/148 because it keeps all of the related code in the config file. That handles the case where the logic isn't too complex. I still like this approach for more complicated scenarios.

PR #150, opened 2015-06-24
Sniff for nonce support

oreoshake:
Because safari pukes when it sees a nonce. it still works (falls back to `unsafe-inline`) but it's unsightly. 

screen shot 2015-06-24 at 11 56 04 am

/cc @maddox


maddox:
:dancer:

PR #151, opened 2015-06-26
Fix missing comma in the recommended initializer

vanderhoorn:


oreoshake:
Thanks! Good :eyes: and/or sorry for the exceptions it would have raised!

PR #153, opened 2015-07-02
cover more ruby versions with coveralls, don't error out on 1.8.7

oreoshake:


Gemfile
```diff
@@ -10,6 +10,6 @@ group :test do
   gem 'rspec', '>= 3.1'
   gem 'growl'
   gem 'rb-fsevent'
-  gem 'coveralls', :platforms => [:ruby_19]
+  gem 'coveralls', :platforms => [:ruby_19, :ruby_20, :ruby_21, :ruby_21]
```
reedloden: You mean :ruby_22 at the end here?

oreoshake: Fixed in https://github.com/twitter/secureheaders/commit/547286d0b6653326a25c412df243bee65fcdc81e

PR #159, opened 2015-08-14
Allow user to disable auto data uri whitelist

rbhitchcock:
The current behavior results in a data URI being appended to the ```img-src``` directive regardless of whether or not the user actually wants to allow data URIs. This PR introduced a new configuration option that the user can specify in order to disable the auto-whitelisting.


oreoshake:
Thanks @rbhitchcock, I'm trying to figure out what's going on with travis. Once I get that squared away I'll merge and release this

Issue #170, opened 2015-09-25
Some CSP directives that appear to be supported are not documented


nchelluri:
E.g.: Current README.md: https://github.com/twitter/secureheaders/blob/140d29eaffb30a970162c50473d59ddff37a874d/README.md#content-security-policy-csp

This file does not mention the `child-src` CSP directive. I believe it is in fact supported, though, because of this bit of code: https://github.com/twitter/secureheaders/blob/a2664d32f458e2d1119c0bfd04300376993a66ee/lib/secure_headers/headers/content_security_policy.rb#L26

I believe that in fact none of the directives listed in `NON_DEFAULT_SOURCES` are documented.

nchelluri:
I think that `child_src` should actually be part of `DIRECTIVES`, since according to https://developer.mozilla.org/en-US/docs/Web/Security/CSP/CSP_policy_directives#default-src it falls back to the `default_src` directive. But the truth is I've just started reading about this stuff recently and I'm not really sure. I also think that `child-src` is a CSP Level 2 directive (and I believe that the most recent browsers supporting CSP at all support level 2, but again, not sure) and that might explain why it wasn't there to begin with.

Furthermore I believe that the default CSP should change it's `frame-src` directive to `child-src` since I believe that `frame-src` is now deprecated in favor of `child-src`.

See also: http://www.w3.org/TR/CSP2/#directive-child-src

nchelluri:
I believe there's a bit more than I thought to this issue. If I specify a `child_src` directive and don't specify `frame_src`, then `frame_src` will use the `default_src` directive (because of the fill in missing directives thing, I think). Specifying an empty `frame_src` string does not seem to work either; I think that disallows every frame entirely. I was hoping `frame_src: nil` would work but it seems to be the same as if I had not specified the `frame_src` at all (uses `default_src` instead).

oreoshake:
@nchelluri nice analysis. Interesting in making a pull request for at least the undocumented behavior? 

I agree that `child_src` chould move from `NON_DEFAULT_SOURCES` to `DIRECTIVES`.

I'm wondering if the "fill all empty directives" functionality is really useful anymore. The spec says browsers should send an effective-directive value, but it wasn't implemented at the time that functionality was added. Removing this would help clear up any confusion.

oreoshake:
WRT to the filling of empty directives, I just checked chrome/firefox (nightly) to see if the `effective-directive` is sent but only Chrome sends it. 

Sigh.

[...]

PR #176, opened 2015-10-01
Remove the 'fill missing directives with default-src value' feature

oreoshake:
1. This should not be required since browsers should implement the `effective-directive` portion of the spec.
1. Managing what is considered an inherited value for the various browser versions can get tricky (see https://github.com/twitter/secureheaders/issues/170).
1. It's not that hard to do the same thing manually
1. It somewhat encourages lazy, over permissive policies.


oreoshake:
Since this should be functionally equivalent to the previous behavior, I'm not going to wait for a major version release. Even though the value of the header and the incoming reports may change, it should have the same effect.

PR #179, opened 2015-10-06
Ua sniffing for support

oreoshake:
Integrating the browser sniffing / directive use that GitHub uses internally. Use the UA to determine which directives should be supported. Provides better validation of config values.

### DIFF
--- a/lib/secure_headers/headers/content_security_policy.rb
+++ b/lib/secure_headers/headers/content_security_policy.rb
@@ -67,7 +67,7 @@ module SecureHeaders
         DIRECTIVES_2_0 + DIRECTIVES_DRAFT
       ).freeze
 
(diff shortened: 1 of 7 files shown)


lib/secure_headers/headers/content_security_policy.rb
```diff
@@ -19,20 +20,53 @@ module Constants
...
+        :sandbox,
         :script_src,
         :style_src,
-        :base_uri,
+        :report_uri
+      ].freeze
+
+      DIRECTIVES_2_0 = [
+        DIRECTIVES_1_0,
+        :base_url,
```
reedloden: base_uri, no?

https://developer.mozilla.org/en-US/docs/Web/Security/CSP/CSP_policy_directives#base-uri

oreoshake: definitely, thanks for the :eyes:

Issue #180, opened 2015-10-06
Proposal: New API for dynamic settings, moving closer to rack


oreoshake:
Thought: only `x-frame-options`, `content-security-policy`, and `hpkp` generally cannot be applied site wide without exception. So provide an API for managing these three directly and as an added bonus we will set all headers in Rack. 

This is backwards compatible with the existing APIs for setting per-request values [\[1\]](https://github.com/twitter/secureheaders#per-action-configuration) [\[2\]](https://github.com/twitter/secureheaders/issues/167#issuecomment-141724924), include script hashes and nonce.

These methods can be called statically from anywhere code path that runs per request. The methods will manipulate a configuration object stored in `ENV`. The provided Rack middleware will merge `SecureHeader::header_hash` into the rack headers. By default, the library will read the values out from `ENV` and fall back to globally configured options. This includes headers not mentioned as having dedicated APIs for changing the settings (i.e. if you really want to, modify the ENV for that config, but you really shouldn't). 

```ruby
# Produces: default-src 'self'; script-src 'self'; object-src 'none'
::SecureHeaders::Configuration.configure do |config|
   config.csp = {
     default_src: %w('self'),
     script_src: %w('self')
   }
   config.x_frame_options = "DENY"
   config.hpkp =  {
    :max_age => 60.days.to_i,
    :include_subdomains => true,
    :report_uri => '//example.com/uri-directive',
    :pins => [
      {:sha256 => 'abc'},
      {:sha256 => '123'}
    ]
  }
end

# Append value to the source list, override 'none' values
# Produces: default-src 'self'; script-src 'self' s3.amazaonaws.com; object-src 'self' youtube.com
append_content_security_policy_source(script_src: "s3.amazaonaws.com", object_src: %w('self' youtube.com))

# Overrides the previously set source list, override 'none' values
# Produces: default-src 'self'; script-src s3.amazaonaws.com; object-src 'self'
override_content_security_policy_directive(script_src: "s3.amazaonaws.com", object_src: %w('self'))

override_x_frame_options(SAMEORIGIN)

# Disable the header if set as a global config setting
unless made_up_method_use_hpkp?
  override_hpkp(false)
end

# Or define the hpkp inline (if the global config is nil/false)
if made_up_method_use_hpkp?
  override_hpkp {
    :max_age => 60.days.to_i,
    :include_subdomains => true,
    :report_uri => '//example.com/uri-directive',
    :pins => [
      {:sha256 => 'abc'},
[...]

PR #181, opened 2015-10-14
[WIP] New API Implementation

oreoshake:
This diff is unnecessarily large, probably better to just checkout the source and view it without the diff. 

major changes
====

* Per-request overrides are stored in `request.env` (is `Thread.local` better?), values are determined and set in [rack middleware](https://github.com/twitter/secureheaders/blob/6b1642001b2cb580bc83846d7c660308e025876e/lib/secure_headers/middleware.rb#L10). 
* API available for overriding/managing CSP (`append_content_security_policy_exception, override_content_security_policy_directives`, `content_security_policy_nonce`), XFO (`override_x_frame_options`), and HPKP (`override_hpkp`) per request, unofficial API for overriding the rest. All are instance methods on a controller object and have `SecureHeaders` class methods too.
* Opting out is available with `SecureHeaders::opt_out_of(SecureHeaders::<HEADERNAME>::CONFIG_KEY)`

Under the hood changes
====
* Only CSP and HPKP are instantiated as objects, all other headers will be configured Strings or defaults. This should keep the number of created objects small, hopefully.
* Configs are validated at boot time, or per request if RAILS_ENV is dev/test (controversial)

Config changes
====
* Only CSP and HPKP support hash arguments. All other values must be simple strings
* CSP directive values must be arrays of strings (not space-delimited values)
* `false` does not mean opt-out. `SecureHeaders::OPT_OUT` does.
* Old values such as `self` `none` `inline` `eval` will raise exceptions when validated. They are no longer valid and the standard values should be used `'none'`, `'self'`, `'unsafe-inline'`, `'unsafe-eval'`

TODO
===
* [x] Add tests for the case when secureheaders is mixed in to a class (rather than statically calling it each time)
* [x] Handle nonces and precomputed headers
* [x] Ensure the README is still valid (it probably isn't)
* [x] Clean this shit up
* [x] Handle overriding of default CSP configs (i.e. CSP config isn't supplied but override/append is called)
* [x] Cache [default headers](https://github.com/twitter/secureheaders/blob/931a32491fc42545cb1cce37b35336f583ebbb40/lib/secure_headers.rb#L53)
* [x] Cache CSP headers per browser (Firefox, Chrome/Opera, Safari, other)
* [x] Add "precomputing" of values concept per [@mastahyeti](https://github.com/twitter/secureheaders/issues/180#issuecomment-146224875)
* [ ] Re-add script hash support (ripped it out since this is a re-write and it was slowing me down)

/cc @mastahyeti @ptoomey3 @gregose


lib/secure_headers/headers/public_key_pins.rb
```diff
@@ -1,23 +1,37 @@
...
-      HPKP_HEADER_NAME = "Public-Key-Pins"
-      ENV_KEY = 'secure_headers.public_key_pins'
-      HASH_ALGORITHMS = [:sha256]
-      DIRECTIVES = [:max_age]
-      CONFIG_KEY = :hpkp
-    end
+    HEADER_NAME = "Public-Key-Pins".freeze
+    HASH_ALGORITHMS = [:sha256].freeze
+    DIRECTIVES = [:max_age].freeze
+    CONFIG_KEY = :hpkp
```
mastahyeti: Since we're using these in the `request.env`, it seems like they should be more unique to avoid conflicts. Maybe `SECURE_HEADERS_HPKP`, for example?

lib/secure_headers.rb
```diff
@@ -30,14 +39,60 @@ module SecureHeaders
   module Configuration
     class << self
```
mastahyeti: It seems a bit weird to me to be using class-level instance variables so much. That seems to be how it was before this PR, but at some point do you think it would be worthwhile to to refactor to have a Configuration instance that is passed around?

lib/secure_headers.rb
```diff
@@ -30,14 +39,60 @@ module SecureHeaders
...
+      end
+
+      private
+      def generate_default_headers
+        default_headers = {}
+        # generate defaults for the "easy" headers
+        (ALL_HEADER_CLASSES - [SecureHeaders::CSP]).each do |klass|
+          config = fetch(klass::CONFIG_KEY)
+          unless config == SecureHeaders::OPT_OUT
+            default_headers[klass::CONFIG_KEY] = klass.make_header(config)
```
mastahyeti: Wouldn't hurt to freeze the results of `make_header(config)` too.

lib/secure_headers.rb
```diff
@@ -30,14 +39,60 @@ module SecureHeaders
...
+          end
+        end
+
+        unless self.csp == SecureHeaders::OPT_OUT
+          default_headers[SecureHeaders::CSP::CONFIG_KEY] = {}
+
+          SecureHeaders::CSP::VARIATIONS.each do |name, _|
+            csp_config = fetch(SecureHeaders::CSP::CONFIG_KEY)
+            csp = SecureHeaders::CSP.make_header(csp_config, UserAgentParser::UserAgent.new(name))
+            default_headers[SecureHeaders::CSP::CONFIG_KEY][name] = csp
```
mastahyeti: ditto here.

lib/secure_headers.rb
```diff
@@ -45,181 +100,158 @@ def configure &block
...
-          options[klass::Constants::CONFIG_KEY]
+        header_config = request.env[klass::CONFIG_KEY] ||
+          secure_headers_request_config(request)[klass::CONFIG_KEY]
+
+        header_name, header = if header_config
+          make_header(klass, header_config, request.user_agent)
         else
-          ::SecureHeaders::Configuration.send(klass::Constants::CONFIG_KEY)
+          # use the cached default, if available
+          if default_header = SecureHeaders::Configuration::default_headers[klass::CONFIG_KEY]
```
mastahyeti: This is a weird syntax for accessing a method. How about `SecureHeaders::Configuration.default_headers`, unless you think `::` is more idiomatic.

lib/secure_headers.rb
```diff
@@ -45,181 +100,158 @@ def configure &block
...
 
-    def header_hash(options = nil)
+    # Strips out headers not applicable to this request
+    def header_hash_for(request)
+      unless request.scheme == HTTPS
+        secure_headers_request_config(request)[SecureHeaders::StrictTransportSecurity::CONFIG_KEY] = SecureHeaders::OPT_OUT
+        secure_headers_request_config(request)[SecureHeaders::PublicKeyPins::CONFIG_KEY] = SecureHeaders::OPT_OUT
+      end
+
       ALL_HEADER_CLASSES.inject({}) do |memo, klass|
```
mastahyeti: Using `each_with_object` instead of `inject` would save you a line on [L140](https://github.com/twitter/secureheaders/pull/181/files#diff-e7fcd10663004caed933b95e0ff2e643L65).

mastahyeti: This block is pretty complex and hard to read. Do you think any of the logic could be extracted into helper methods?

lib/secure_headers.rb
```diff
@@ -45,181 +100,158 @@ def configure &block
...
-        config = if options.is_a?(Hash) && options[klass::Constants::CONFIG_KEY]
-          options[klass::Constants::CONFIG_KEY]
+        header_config = request.env[klass::CONFIG_KEY] ||
+          secure_headers_request_config(request)[klass::CONFIG_KEY]
```
mastahyeti: Are these two separate API's for overriding the config per-request? If so, it would be nice to just pick one to go with for simplicity's sake.

lib/secure_headers.rb
```diff
@@ -45,181 +100,158 @@ def configure &block
...
+        unless config[name]
+          config[name] = config[:default_src]
         end
+      end
 
-        ActiveSupport::Notifications.subscribe("render_template.action_view") do |event_name, start_at, end_at, id, payload|
-          save_hash_for_later payload
-        end
+      secure_headers_request_config(request)[SecureHeaders::CSP::CONFIG_KEY] = config.merge(additions) do |_, lhs, rhs|
+        lhs | rhs
```
mastahyeti: A comment here would be good. I don't understand this.

lib/secure_headers.rb
oreoshake: good call.

lib/secure_headers.rb
oreoshake: Agreed. I think the code is trying to be too generic when there's many possible cases.

lib/secure_headers/headers/public_key_pins.rb
oreoshake: All of these CONFIG_KEYS will be scoped to https://github.com/twitter/secureheaders/pull/181/files#diff-e7fcd10663004caed933b95e0ff2e643R23 in https://github.com/twitter/secureheaders/pull/181/files#diff-e7fcd10663004caed933b95e0ff2e643R160. That will be "unsupported" when I remove the line from your comment about pulling the config from two different places in `request.env`. 

Unfortunately that's a left over from when I thought passing a hash directly to `header_hash_for` would be a thing. :hankey:

lib/secure_headers.rb
oreoshake: Yeah, that will also make it easier to scope the configuration for the given precomputed overrides.

oreoshake:
This PR is entirely too large. Having responded to @mastahyeti's comments I'd say it's pretty unusable. I'm going to create a 3.x branch, commit the mass deletion of the fixtures directory, and create a new PR based on the 3.x branch. It will somewhat shorten this (although I still think it's better to look at the code as new and ignore the diff in general)

lib/secure_headers.rb
oreoshake: Re `each_with_object` zomg I love this post ruby 1.8.7 world I now live in.

oreoshake:
clsoing in favor of https://github.com/twitter/secureheaders/pull/191

PR #191, opened 2015-11-05
New API Implementation

oreoshake:
Followup to https://github.com/twitter/secureheaders/pull/181. As with the last PR, it's probably better to ignore the diff and look at the code directly.

The biggest change in this PR is the added support for "named overrides" - additional configuration objects that can be referenced by name. The header values are all precomputed so named overrides save the overhead of building a policy per request.

(from [the readme](https://github.com/twitter/secureheaders/blob/b0991eed620eacf4ddd5e94d1b788885bb3b1d60/README.md#named-overrides)

```ruby
class ApplicationController < ActionController::Base
    config.csp = {
      default_src: %w('self'),
      script_src: %w(example.org)
    }
  end

  # override default configuration
  SecureHeaders::Configuration.override(:script_from_otherdomain_com) do |config|
    config.csp[:script_src] << "otherdomain.com"
  end

  # overrides the :script_from_otherdomain_com configuration
  SecureHeaders::Configuration.override(:another_config, :script_from_otherdomain_com) do |config|
    config.csp[:script_src] << "evenanotherdomain.com"
  end
end

class MyController < ApplicationController
  def index
    # Produces default-src 'self'; script-src example.org otherdomain.org
    use_secure_headers_override(:script_from_otherdomain_com)
  end

  def show
    # Produces default-src 'self'; script-src example.org otherdomain.org evenanotherdomain.com
    use_secure_headers_override(:another_config)
  end
end
```


README.md
```diff
@@ -10,243 +10,197 @@ The gem will automatically apply several headers that are related to security.
...
+    :frame_ancestors => %w('none'),
+    :plugin_types => %w(application/x-shockwave-flash),
+    :block_all_mixed_content => true # see [http://www.w3.org/TR/mixed-content/](http://www.w3.org/TR/mixed-content/)
+    :report_uri => %w(https://example.com/uri-directive)
   }
   config.hpkp = {
     :max_age => 60.days.to_i,
     :include_subdomains => true,
-    :report_uri => '//example.com/uri-directive',
+    :report_uri => "https//example.com/uri-directive",
```
stve: missing a `:` here

oreoshake:
This code is now running on `github.com`. I'll let it sit for a few days and release a 3.0 gem. I'll add an "upgrading to 3.0 wiki" entry in the meantime since it is very much a breaking change.

PR #197, opened 2015-12-11
Use clean ruby hash syntax

mnoack:
* Ruby 1.8 has not been supported for years


oreoshake:
:metal: 

I'm actually surprised there wasn't more use of the old syntax.

PR #199, opened 2015-12-15
Specify  "max-age" in HSTS string to prevent error.

kplattret:


oreoshake:
Ah, thanks. I wish my documentation was runnable.

PR #201, opened 2015-12-22
added `to_h` to nonced_tag in view helper

xssc:
This will make it so you do not need to supply options for it to work, just setting it to nil or a blank string causes a error when you try to merge the nonce.


oreoshake:
@xssc thanks for the bug report. I haven't deployed the 3.x branch with nonces yet :disappointed:. 

What do you think about isntead setting `content_or_options = {}` for the two `nonced_*_tag` method signatures? 

I really with this class had spec tests too.

xssc:
Thought about that to, makes more sense

PR #215, opened 2016-02-18
2.x maintenance: Use before_action if possible, for Rails 5 support

chrisarcand:
This should allow use of Rails 5 without deprecations from `before_filter`.
`before_filter` is deprecated and will be removed in Rails 5.1.

I know the use of before_* at all has been removed from the 3.x release, but as that's not stable yet I think this is a good way to go.

### DIFF
--- a/lib/secure_headers.rb
+++ b/lib/secure_headers.rb
@@ -27,6 +27,18 @@ module SecureHeaders
     SecureHeaders::XXssProtection
   ]
 
+  ALL_FILTER_METHODS = [
+    :prep_script_hash,
+    :set_hsts_header,
+    :set_hpkp_header,
+    :set_x_frame_options_header,
+    :set_csp_header,
+    :set_x_xss_protection_header,
+    :set_x_content_type_options_header,
+    :set_x_download_options_header,
+    :set_x_permitted_cross_domain_policies_header
+  ]
+
   module Configuration
     class << self
       attr_accessor :hsts, :x_frame_options, :x_content_type_options,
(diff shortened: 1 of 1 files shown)


lib/secure_headers.rb
```diff
@@ -97,15 +97,19 @@ def ensure_security_headers options = {}
...
-      before_filter :prep_script_hash
-      before_filter :set_hsts_header
-      before_filter :set_hpkp_header
-      before_filter :set_x_frame_options_header
-      before_filter :set_csp_header
-      before_filter :set_x_xss_protection_header
-      before_filter :set_x_content_type_options_header
-      before_filter :set_x_download_options_header
-      before_filter :set_x_permitted_cross_domain_policies_header
+      actions = [:prep_script_hash,
```
oreoshake: Much nicer! However, this should probably be a constant.

oreoshake:
Rails 5... must be nice :smile:

lib/secure_headers.rb
chrisarcand: Derp. Late night. Thanks :smile:

lib/secure_headers.rb
```diff
@@ -97,15 +97,10 @@ def ensure_security_headers options = {}
...
-      before_filter :set_x_frame_options_header
-      before_filter :set_csp_header
-      before_filter :set_x_xss_protection_header
-      before_filter :set_x_content_type_options_header
-      before_filter :set_x_download_options_header
-      before_filter :set_x_permitted_cross_domain_policies_header
+      hook = respond_to?(:before_action) ? :before_action : :before_filter
+      [:prep_script_hash, :set_security_headers].each do |method|
+        send(hook, method)
+      end
```
chrisarcand: @oreoshake Actually, I just noticed that all of these methods are included in `set_security_headers` (except `prep_script_hash`) How's this, instead?

oreoshake: The reason for setting the different `before_filter`s is that it makes it possible to use `skip_before_filter` for each individual header. Calling `set_security_headers` would remove this granularity.

chrisarcand: Oooh of course. Thanks for the clarification ^_^

oreoshake:
Just pushed 2.5.2 with this change.

merci beaucoup!

chrisarcand:
Much appreciated, thanks @oreoshake! :beers:

PR #222, opened 2016-02-28
Remove ensure_security_headers for 3.x upgrade

mattbrictson:
I think the 2.x → 3.x upgrade instructions are missing a step: "remove the `ensure_security_headers` from your controllers."


oreoshake:
Thanks! This certainly is an important step if you want your app to boot!

PR #224, opened 2016-02-29
Handle policy merges when directives do not inherit from default-src

oreoshake:
Previously, non default source values would be merged with the default-src value which makes no sense. This was especially confusing if your default-src was `*` or `'none'`

I'm going to wait until I confirm that `plugin-types` is a non-default source directive. The spec does not mention it explicitly, and I've submitted https://github.com/w3c/webappsec/pull/510 in case my hunch is correct.


oreoshake:
Confirmed, `plugin-types` is a non-default-src directive. More clearly, `default-src` only applies to things governed by fetch.

PR #229, opened 2016-03-14
Fix missing ``` in changelog

jgeiger:
Fix changelog formatting


oreoshake:
Good :eyes:. Please move the triple \` up one more line so 

> Would result in `default-src *; report-uri *` which doesn't make any sense at all.

isn't included in that code block.

Issue #145, opened 2015-06-10
No way to set the secure option for Set-Cookie header


jmera:
I'm not sure if this is in the works.  It'd be nice to be able to easily set the Secure option on the `Set-Cookie` header so it looks something like:

<code>Set-Cookie: _app_name=some_string; path=/; expires=Wed, 17-Jun-2015 20:41:13 GMT; HttpOnly; <b>Secure</b></code>

Info about the flag here: https://www.owasp.org/index.php/SecureFlag

oreoshake:
I agree: a "mark all cookies as secure" feature. I have to think something else already does this, perhaps in a rack middleware.

jmera:
You're right... https://github.com/tobmatth/rack-ssl-enforcer already does this + more.  Nonetheless, for those wanting a one-gem solution, do you think this might be a useful feature to add to secure_headers?  I'd be willing to open up a PR to try to get this working.

oreoshake:
Certainly!

oreoshake:
@jmera any interest in continuing with this? We can lift https://github.com/tobmatth/rack-ssl-enforcer/blob/master/lib/rack/ssl-enforcer.rb#L183 as I can't see any way to improve upon what was already implemented (other than extracting that regex to a constant)

jmera:
@oreoshake I'll give it a shot! I already wrote something similar a while back: https://github.com/jmera/secure_cookies/blob/master/lib/rack/secure_cookies.rb#L20

Will open up a PR :smile:

jmera:
@oreoshake the only way I see this working is through Rack middleware that lies on top of the app's cookie middleware (so we can have access to `headers['Set-Cookie']`). Essentially I could just merge in `jmera/secure_cookies` and include instructions on how to use the middleware like so: https://github.com/jmera/secure_cookies#usage

We could try to make this a config option and "automagically" place said middleware at the top of the stack.

```ruby
::SecureHeaders::Configuration.default do |config|
  config.flag_cookies_as_secure = ENV['RACK_ENV'] == 'production'
  ...
end
```
 At the moment it isn't completely clear to me how we'd accomplish the automagical part...

oreoshake:
You could add it [to the existing middleware](https://github.com/twitter/secureheaders/blob/master/lib/secure_headers/middleware.rb) and have it noop if the flag is false. Your existing documentation will instruct people where to place the middleware (it's also added via a railtie just in case)

jmera:
Implemented in #231

PR #231, opened 2016-03-15
Secure Cookies

jmera:
See #145 for initial discussion

We would like to give users the ability to flag all cookies as secure: https://www.owasp.org/index.php/SecureFlag

### DIFF
--- a/lib/secure_headers/middleware.rb
+++ b/lib/secure_headers/middleware.rb
@@ -1,5 +1,7 @@
 module SecureHeaders
   class Middleware
+    SECURE_COOKIE_REGEXP = /;\s*secure\s*(;|$)/i.freeze
+
     def initialize(app)
       @app = app
     end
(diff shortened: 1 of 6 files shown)


oreoshake:
I'll drop 1.9.3 in the next major release, but for now please make travis happy.

lib/secure_headers/middleware.rb
```diff
@@ -8,8 +10,32 @@ def initialize(app)
...
+            "#{cookie}; secure"
+          else
+            cookie
+          end
+        end.join("\n")
+      end
+    end
+
+    def config(req)
+      req.env[SECURE_HEADERS_CONFIG] || Configuration.get(Configuration::DEFAULT_CONFIG)
```
jmera: @oreoshake is there an easier way to access the config from this context?

oreoshake: `req.env[SECURE_HEADERS_CONFIG]` should suffice.

lib/secure_headers/railtie.rb
```diff
@@ -10,7 +10,7 @@ class Railtie < Rails::Railtie
                              'Public-Key-Pins', 'Public-Key-Pins-Report-Only']
 
       initializer "secure_headers.middleware" do
-        Rails.application.config.middleware.use SecureHeaders::Middleware
+        Rails.application.config.middleware.insert_before 0, SecureHeaders::Middleware
```
jmera: For this feature to work correctly `SecureHeaders::Middleware` needs to come before the session middleware (or simply at the top of the rack stack).

lib/secure_headers/middleware.rb
oreoshake: Or maybe make `config_for` public? https://github.com/twitter/secureheaders/blob/44c8c3edfe6c5456274c448ba68371b20fe649d7/lib/secure_headers.rb#L225

lib/secure_headers/railtie.rb
oreoshake: This should be fine.

lib/secure_headers/middleware.rb
jmera: Sounds good to me, will help keep things dry

oreoshake:
Woot. I'll release this with https://github.com/twitter/secureheaders/pull/232 tomorrow.

jmera:
:beers:

PR #232, opened 2016-03-15
Handle very dynamic policies better

### LINKED ISSUE #217: [3.x] Only perform idempotency check once per request
As of now, any call to `append/override_content_security_policy_directives` will perform a check to see whether or not the changes will affect the underlying config. Mismatches in config cause the pre-generated header cache to be erased.

This [idempotency check](https://github.com/twitter/secureheaders/blob/master/lib/secure_headers/headers/content_security_policy.rb#L187) is not trivial and creates a few temporary objects. If you have many calls to `append/override_content_security_policy_directives` in one request cycle, then you are likely making a lot of unnecessary checks when all you really care about is if the resulting policy actually changed.

Storing the modified config object and performing the idempotent check one per request should have the same effect, but will be more performant and use less memory.

Splittin' hairs here, but is definitely an improvement

oreoshake:
Fixes #217 

Currently, an idempotency check is performed for every policy addition. Not only does this create a fair number of temporary objects, but it also busts the entire header cache if any changes are detected.

This change defers idempotency checks until just before the header values are generated. If we can determine that the CSP will differ from the cached values, we discard all cached CSP and generate one specific to that UA.

If there is a change to `X-Frame-Options`, we cache the new value immediately.

As a result, the value for `cached_headers` will always maintain the headers that will be applied and there is no need to generate them when calling `header_hash_for`. 

Because we are retaining the previously discarded cached headers, we only call `dup` on the configuration option if it is `frozen` to save bytes and cycles. All policies added via `default` and `override` are already frozen automatically, so a non-`frozen` config object can be modified freely and we can assume it's dynamic in nature.

To recap:

| Before | After |
| --- | --- |
| `idempotent_additions?` called per policy modification | `idempotent_additions?` called once per request |
| Entire header cache was busted per policy/xfo modification | the CSP header cache is busted once per request, only policies for the current UA are regenerated |
| config objects were `dup`'d per modification | config values `dup`'d at most one time per request |
| XFO changes busted the entire header cache | XFO changes modify the header cache directly |
| Opting out of a header would bust the cache | Opting out of a header modifies the cache |

### DIFF
--- a/spec/lib/secure_headers/configuration_spec.rb
+++ b/spec/lib/secure_headers/configuration_spec.rb
@@ -29,16 +29,14 @@ module SecureHeaders
       expect_default_values(header_hash)
     end
 
-    it "copies all config values except for the cached headers when dup" do
+    it "copies config values when duping" do
       Configuration.override(:test_override, Configuration::NOOP_CONFIGURATION) do
         # do nothing, just copy it
       end
 
       config = Configuration.get(:test_override)
       noop = Configuration.get(Configuration::NOOP_CONFIGURATION)
-      [:hsts, :x_frame_options, :x_content_type_options, :x_xss_protection,
-        :x_download_options, :x_permitted_cross_domain_policies, :hpkp, :csp, :secure_cookies].each do |key|
-
+      [:csp, :dynamic_csp].each do |key|
         expect(config.send(key)).to eq(noop.send(key)), "Value not copied: #{key}."
       end
     end
--- a/spec/lib/secure_headers/configuration_spec.rb
+++ b/spec/lib/secure_headers/configuration_spec.rb
@@ -36,7 +36,7 @@ module SecureHeaders
 
       config = Configuration.get(:test_override)
       noop = Configuration.get(Configuration::NOOP_CONFIGURATION)
-      [:csp, :dynamic_csp].each do |key|
+      [:csp, :dynamic_csp, :secure_cookies].each do |key|
         expect(config.send(key)).to eq(noop.send(key)), "Value not copied: #{key}."
       end
     end
--- a/lib/secure_headers.rb
+++ b/lib/secure_headers.rb
@@ -75,8 +75,7 @@ module SecureHeaders
     # Returns the current config
     def override_x_frame_options(request, value)
       config = config_for(request)
-      config.x_frame_options = value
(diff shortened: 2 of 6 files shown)


oreoshake:
I suppose with this change you could argue that we only need to dup/operate on the header cache instead of maintaining a parallel config. Only the case of CSP do we operate on the config values without immediately modifying the header cache. 

The only reason to keep the config fields around is to lookup the value (e.g. the CSP config hash) and I'm not sure this would ever happen IRL, or at the least it would be an antipattern - business logic around what headers are currently configured should probably be handled in application code.

oreoshake:
This is ready for some :eyes: if anyone has a moment. 

Note: this is a breaking API change - retrieving config values will only work for CSP and secure_cookies. I really can't see anyone working directly with a config object for the other headers and trying to retrieve their values. e.g.

:boom:
```ruby
SecureHeaders.override(:my_override) do |config|
  config.x_frame_options = config.x_frame_options == "sameorigin" ? "sameorigin" : "deny"
end
```

No :boom:
```ruby
SecureHeaders.override(:my_override) do |config|
  config.csp = config.csp.merge(:script_src => %w(data:)) # don't ever add data: to scriptsrc btw
end
```

spec/lib/secure_headers/configuration_spec.rb
```diff
@@ -29,16 +29,14 @@ module SecureHeaders
       expect_default_values(header_hash)
     end
 
-    it "copies all config values except for the cached headers when dup" do
+    it "copies config values when duping" do
```
ptoomey3: Given this PR is optimizing policy overrides are, I was surprised there was no test to validate that you are in fact only busting the cache once even if there are > 1 overrides. Does that seem a reasonable thing to add?

ptoomey3: Would it also make sense to add a test to verify that overrides never end up touching `csp` and only touch `dynamic_csp`?

ptoomey3:
Other than small test nits, this seems logical. :+1:

spec/lib/secure_headers/configuration_spec.rb
oreoshake: Added a test [here](https://github.com/twitter/secureheaders/pull/232/files#diff-5ffee118dcfe384935df780237dadf39R109) for the cache busting. I'll have to think about your second item.

PR #238, opened 2016-03-30
add cookies configuration

stve:
Add configuration and support for:

* [x]  `Secure` cookies
* [x]  `HttpOnly` cookies
* [x] `SameSite` cookies
* [x] `except`/`only` configuration for cookie whitelisting
* [x] cookie configuration validation (don't allow simultaneous use of `except` and `only`)
* [x] deprecate `secure_cookies` configuration
* [x] documentation


oreoshake:
Great start!

stve:
Thanks @oreoshake, I want to refactor it a bit once it's in place. I'm not really happy with the complexity in the `flag_cookie!` method right now.

stve:
@oreoshake I'd be curious to hear your thoughts on 4aaac0ce242d31459a4dc7d602070c085a6f2f81 before I continue to flesh out the SameSite cookie support. I felt like the code in `Middleware` was getting a bit bloaty. I contemplated adding a middleware exclusively for cookie manipulation, but that seemed excessive, so encapsulating the logic inside a `Cookie` class was the next logical step.

I left the tests for cookie headers in the `Middleware` spec for the time being, but I'm thinking they may be redundant since the `Cookie` tests more or less provide the same assertions.

lib/secure_headers/cookie.rb
```diff
@@ -0,0 +1,78 @@
...
+    def to_s
+      @raw_cookie.dup.tap do |c|
+        c << "; secure" if secure?
+        c << "; HttpOnly" if httponly?
+        c << "; SameSite" if samesite?
+      end
+    end
+
+    def secure?
+      already_flagged?(:secure) || flag_cookie?(:secure)
```
oreoshake: I think this would be better if the conditions were swapped so we can short circuit on a hash lookup instead of a regex match.

oreoshake:
I think it's :ok: to leave it in one middleware for now.

lib/secure_headers/middleware.rb
```diff
@@ -12,7 +10,8 @@ def call(env)
       status, headers, response = @app.call(env)
 
       config = SecureHeaders.config_for(req)
-      flag_cookies_as_secure!(headers) if config.secure_cookies
+      flag_cookies!(headers, config.cookies) if config.cookies
+      flag_cookies!(headers, secure: true) if config.secure_cookies
```
oreoshake: Please add a deprecation warning here. I know it's brand new code but I plan to get rid of it and use this format.

stve: Absolutely. I was going to ask you about that. I was thinking I'd add it on configuration

stve:
@oreoshake I think this is at a point where it is ready for review. There are probably some edge cases on the validations that might need more test coverage (as I was writing this it occurred to me that the SameSite config should prevent a cookie from being configured as both Lax and Strict)

README.md
```diff
@@ -264,6 +268,55 @@ config.hpkp = {
...
+}
+```
+
+#### Hash-based configuration
+
+Hash-based configuration allows for fine-grained control.
+
+```ruby
+config.cookies = {
+  secure: { except: ['_guest'], # mark all but the `_guest` cookie as Secure
```
oreoshake: missing closing `}`

lib/secure_headers/configuration.rb
```diff
@@ -164,13 +164,19 @@ def validate_config!
       XDownloadOptions.validate_config!(@x_download_options)
       XPermittedCrossDomainPolicies.validate_config!(@x_permitted_cross_domain_policies)
       PublicKeyPins.validate_config!(@hpkp)
+      Cookie.validate_config!(@cookies)
+    end
+
+    def secure_cookies=(secure_cookies)
```
oreoshake: :+1:

lib/secure_headers/headers/cookie.rb
```diff
@@ -0,0 +1,136 @@
...
+        true
+      else
+        false
+      end
+    end
+
+    def samesite_cookie
+      case config[:samesite]
+      when TrueClass
+        "SameSite"
```
oreoshake: is this valid? I'm asking because I have no idea if that means the same thing as `SameSite=Strict`

oreoshake:
This looks pretty good. I'd like to see a contrived test for the combination of lax/strict cookies at the middleware level. Something like:

```ruby 
config.cookies = {
  samesite: {
    strict: { only: [`_rails_session`] },
    lax: { except: [`_rails_session`] }
  }
}

request = Rack::Request.new("HTTPS" => "on")
_, env = cookie_middleware.call request.env
expect(env['Set-Cookie']).to include("_rails_session; SameSite=Strict")
expect(env['Set-Cookie']).to include("_guest; SameSite=Lax")
```

lib/secure_headers/headers/cookie.rb
```diff
@@ -0,0 +1,136 @@
...
+    end
+
+    private
+
+    def parsed_cookie
+      @parsed_cookie ||= CGI::Cookie.parse(raw_cookie)
+    end
+
+    def already_flagged?(attribute)
+      raw_cookie =~ REGEXES[attribute]
```
oreoshake: I wonder if it would be better to parse out the flags into fields so we can use accessors instead of regex to test if the value needs to be modified.

README.md
stve: nice catch, thanks!

lib/secure_headers/headers/cookie.rb
stve: That's a good point. `CGI::Cookie` does some of this already, but unfortunately falls a little bit short since it doesn't support attributes other than `Secure` and `HttpOnly`. I'll take a stab at parsing out the cookie.

lib/secure_headers/headers/cookie.rb
stve: My first read of the spec led me to believe this was a shortcut for `Strict` but now i'm not so sure. Here's the [relevant section](https://tools.ietf.org/html/draft-west-first-party-cookies-07#section-4.1) in the spec:

>   If the "attribute-name" case-insensitively matches the string "SameSite", the user agent MUST process the "cookie-av" as follows:
>   1.  If "cookie-av"'s "attribute-value" is not a case-sensitive match for "Strict" or "Lax", ignore the "cookie-av".
>   2.  Let "enforcement" be "Lax" if "cookie-av"'s "attribute-value" is a case-insensitive match for "Lax", and "Strict" otherwise.
>   3.  Append an attribute to the "cookie-attribute-list" with an "attribute-name" of "SameSite" and an "attribute-value" of "enforcement".

Now that I've re-read it, my interpretation of that would be that rule 1 would apply and it would be ignored. Is that your understanding as well? If so, `samesite: true` shouldn't be valid and I'll remove it.

oreoshake: > Now that I've re-read it, my interpretation of that would be that rule 1 would apply and it would be ignored.

That's my understanding as well.

>  If so, samesite: true shouldn't be valid and I'll remove it.

:+1: 

Thanks!

stve: Do you think it's worth supporting booleans on Strict/Lax? Like:

```ruby
config.cookies = {
  samesite: {
    strict: true
  }
}
```

Use of a boolean on either would prevent a rule of any sort for the other enforcement mode. This could be checked via validation.

oreoshake: Yes. Also, I think the sample you provided would be the ideal setup that people would want to work towards.

stve: :+1:

lib/secure_headers/headers/cookie.rb
stve: I ripped some code out of `CGI::Cookie` and implemented a rudimentary parser so that we can avoid regex testing (2e74b8120a75144e952afe06ebe7ad12d1c98e5f). I like this approach better, thanks for the suggestion.

stve:
@oreoshake is there anything else you'd like to see addressed here?

README.md
```diff
@@ -31,7 +31,11 @@ All `nil` values will fallback to their default values. `SecureHeaders::OPT_OUT`
 
 ```ruby
-  config.secure_cookies = true # mark all cookies as "secure"
+  config.cookies = {
+    secure: true, # mark all cookies as "secure"
+    httponly: true, # mark all cookies as "httponly"
+    samesite: true
```
oreoshake: This isn't valid anymore, right?

lib/secure_headers/headers/cookie.rb
```diff
@@ -0,0 +1,187 @@
...
+      end
+    end
+
+    attr_reader :raw_cookie, :config
+
+    def initialize(cookie, config)
+      @raw_cookie = cookie
+      @config = config
+      @attributes = {
+        "secure" => nil,
```
oreoshake: use symbols here?

oreoshake:
Sorry I missed the earlier ping! Two pieces of feedback:

1. That validation method is pretty hairy. I think it would benefit from being split up into multiple, smaller methods for clarity.
1. I've been debating this internally for no good reason, but I think the tests would benefit from using actual string comparisons instead of regex matches. e.g. [this test](https://github.com/twitter/secureheaders/pull/238/files#diff-1d59181f0c73149a39a236ec1e703aadR13) would pass if the output was `_session=thisisatest; securesecuresecurebatteryhorsestaple`. Because we're just manipulating strings, I think we should test the value and not the format of the output.

Thanks again for all of this :heart: I can't wait to release it!

README.md
stve: that's correct, `samesite` will only accept a Hash, I'll fix this

lib/secure_headers/headers/cookie.rb
stve: We can certainly do that, my reasoning for using strings was that parsing the cookie in order to decompose it would require `to_sym` on strings in order to set these to a truthy value. I guess now that Ruby garbage collects symbols this is less of a concern.

stve:
1. Absolutely, the validation method is cringe worthy, I will try to think through some improvements here.
2. I'm definitely :+1: on moving away from regex matches, this will be a good change.

README.md
```diff
@@ -264,6 +270,57 @@ config.hpkp = {
...
+  }
+}
+```
+
+`Strict` and `Lax` enforcement modes can also be specified using a Hash.
+
+```ruby
+config.cookies = {
+  samesite: {
+    strict: { only: [`_rails_session`] },
```
oreoshake: Looks like you used \` instead of ' so copy pasting failed :smile:

oreoshake:
Tested this out locally and it's looking good. Just need to update the docs and split up the validation method.

README.md
stve: Whoops! :banana:  I've been doing some React/JSX stuff lately and this probably was a bad context switch. Fixed now. :smile:

stve:
> Tested this out locally and it's looking good

ok great! I'm going to work through the validation tonight/tomorrow. :sunglasses:

stve:
I think bc2bc7593a6fa4390362e433cc1ea8c819082d03 gets us closer to a more readable validation method. I contemplated breaking them out further into separate classes but that might be overkill. This definitely feels like an improvement. :smile:

oreoshake:
[This looks like the only spot](https://github.com/twitter/secureheaders/pull/238/files#diff-21eb15c7cc19b93eab6709eeb9e4bab0R60) with room for improvement. I think this is good to go as is, but it would be ✨  if you could remove the near-duplication on those lines.

stve:
Yep, that was easily done :smiley_cat: fc8a80b6aa66366d0529076d5ecc6ade3d6b1155

stve:
Thanks for merging! :heart_eyes:

oreoshake:
I want to release this with https://github.com/twitter/secureheaders/pull/240, so I'll probably cut a new gem soon.

PR #241, opened 2016-03-31
Account for `ActionDispatch::SSL` in middleware ordering

### LINKED ISSUE #237: Secure Headers' config.hsts overriden by Rails' config.force_ssl
I have added Secure Headers to my webapp (for CSP), and so far it has worked great.

Now I am trying to implement HSTS preloading, and added `config.hsts = "max-age=#{20.years.to_i}; includeSubdomains; preload"`, but noticed that Rails still sends a different "Strict-Transport-Security" header, which misses the "preload" keyword.

Upon investigating, it appears that because my Rails environment includes `config.force_ssl = true` (in order to redirect http->https, mark cookies as secure, etc.), Rails overwrites the "Strict-Transport-Security" header. The relevant code (for Rails 4.2) is here:

https://github.com/rails/rails/blob/4-2-stable/actionpack/lib/action_dispatch/middleware/ssl.rb#L48

Is it possible for Secure Headers to take priority over that other middleware?

jacobbednarz:
Rails 4.0 stable [introduced](https://github.com/rails/rails/commit/9ec63eb0491a1b72381833478398c369ab48019a#diff-8e5f6b33c191ad6dec07f3288345a13f) `ActionDispatch::SSL` to replace the previous
dependency on `Rack::SSL` for setting HSTS HTTP headers. In doing so they have
caused an issue with the middleware load order where the values from
`SecureHeaders::Middleware` are overridden by `ActionDispatch::SSL` due to it
being loaded afterwards.

This fixes the load issue by checking if `ActionDispatch::SSL` is used and
inserting the secure headers middleware after it instead of Rails determining
the load order. The approach used here should also allow those running older
Rails applications to maintain compatibility without any changes.

Fixes #237.

### DIFF
--- a/lib/secure_headers/railtie.rb
+++ b/lib/secure_headers/railtie.rb
@@ -10,7 +10,11 @@ if defined?(Rails::Railtie)
                              'Public-Key-Pins', 'Public-Key-Pins-Report-Only']
 
       initializer "secure_headers.middleware" do
-        Rails.application.config.middleware.use SecureHeaders::Middleware
+        if defined?(ActionDispatch::SSL)
+          Rails.application.config.middleware.insert_after ActionDispatch::SSL, SecureHeaders::Middleware
+        else
+          Rails.application.config.middleware.insert_before 0, SecureHeaders::Middleware
+        end
       end
 
       initializer "secure_headers.action_controller" do


jacobbednarz:
Tests to come once I have a look at the middleware setup and best approaches for testing `ActionDispatch::SSL`

oreoshake:
If you can't find anything, I think this is OK to add untested.

jacobbednarz:
@encoder do you mind pulling this locally and confirming it works as expected for you? I'm looking to add some testing but most of them look to be quite brittle.

encoder:
I patched railtie.rb on my end with this, but it didn't seem to work. The header is still sent by Rails, though the middleware seems to be applied correctly:

```
$ bundle exec rails c
Loading development environment (Rails 4.2.5.1)
irb(main):001:0> Rails.application.config.middleware
=> #<ActionDispatch::MiddlewareStack:0x0000000576f3b8 @middlewares=[ActionDispatch::SSL, SecureHeaders::Middleware, ...
```

jacobbednarz:
@encoder are you able to remove the patch and instead call `.freeze` on the headers in https://github.com/twitter/secureheaders/blob/master/lib/secure_headers/middleware.rb#L10-L18 and watch for an exception to be raised to confirm where this is getting mutated?

@oreoshake might be able to point us at a better spot to freeze the headers if the above isn't correct.

encoder:
I was able to solve this by updating to the latest version of secure_headers and removing the patch. Previously I was using secure_headers 3.0.0, which seems to always be getting overwritten by Rails (with or without the patch). When upgrading to 3.1.2 while also keeping the patch, the headers still get overwritten. But the stock 3.1.2 (without this patch) seems to be sending the correct headers.

Upon bisecting the code, the last affected version is 3.0.3.

You can thus mark #237 as closed, and discard the patch. Sorry for all the trouble!

jacobbednarz:
Great news @encoder! Happy to close this one off. :smile:

PR #244, opened 2016-04-19
Additional tests

stve:
Just adding some small test cases, I felt that the undefined override test was important since it's part of the public API. The cookies config is covered as a unit test elsewhere but I added the test here for consistency with the other policies. 😄


oreoshake:
Always good to have more tests!

Issue #218, opened 2016-02-25
Add support for 'referrer-policy' header


reedloden:
https://w3c.github.io/webappsec-referrer-policy/ outlines it.

oreoshake:
I thought the directive was dropped? Looks like it's a separate header: https://w3c.github.io/webappsec-referrer-policy/#referrer-policy-header.

reedloden:
Ah, MDN docs are outdated then. :(

https://developer.mozilla.org/en-US/docs/Web/Security/CSP/CSP_policy_directives#referrer
https://developer.mozilla.org/en-US/docs/Web/Security/CSP/CSP_policy_directives#Specifications

Ok, closing this out then.

oreoshake:
Re-opening, adding the `referrer-policy` header is certainly within scope

reedloden:
Fixed thanks to #249

PR #249, opened 2016-04-25
Add support for Referrer Policy header

tomgilligan:
See #218 for initial discussion.  Forgot to check browser support before going ahead with this 😞 Maybe useful at a later date.

### DIFF
--- a/lib/secure_headers/railtie.rb
+++ b/lib/secure_headers/railtie.rb
@@ -7,7 +7,7 @@ if defined?(Rails::Railtie)
                              'X-Permitted-Cross-Domain-Policies', 'X-Download-Options',
                              'X-Content-Type-Options', 'Strict-Transport-Security',
                              'Content-Security-Policy', 'Content-Security-Policy-Report-Only',
-                             'Public-Key-Pins', 'Public-Key-Pins-Report-Only']
+                             'Public-Key-Pins', 'Public-Key-Pins-Report-Only, Referrer-Policy']
 
       initializer "secure_headers.middleware" do
         Rails.application.config.middleware.insert_before 0, SecureHeaders::Middleware
--- a/lib/secure_headers/railtie.rb
+++ b/lib/secure_headers/railtie.rb
@@ -7,7 +7,7 @@ if defined?(Rails::Railtie)
                              'X-Permitted-Cross-Domain-Policies', 'X-Download-Options',
                              'X-Content-Type-Options', 'Strict-Transport-Security',
                              'Content-Security-Policy', 'Content-Security-Policy-Report-Only',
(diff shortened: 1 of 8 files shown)


oreoshake:
1. Thank you! This is great. :heart:
1. We can't apply this by default without a major version bump, in the meantime let's opt-out by default [like we do for hpkp here](https://github.com/twitter/secureheaders/blob/master/lib/secure_headers/configuration.rb#L119). Even though I believe that 95% of web apps can default to this setting without issue, we just can't surprise that other 5% with broken behavior.
1. @mikewest is the `referrer-policy` header still a thing? I don't see it anywhere under ["referrer policy delivery"](https://www.w3.org/TR/referrer-policy/#referrer-policy-delivery)

mikewest:
Deferring to @estark37 and @jeisinger, who are much more up to date on Referrer Policy than I am.

estark37:
Referrer-Policy is a separate header, but isn't implemented in Chrome yet (and I don't think in FF either). We'll likely deprecate the referrer directive in CSP at some point after implementing the separate header.

oreoshake:
@tomgilligan sounds like `referrer-policy` will be a thing so if you can address # 2 in [my previous comment](https://github.com/twitter/secureheaders/pull/249#issuecomment-214468134) we can merge this to be ready for the browser implementations.

thanks @mikewest and @estark37 😻

lib/secure_headers/railtie.rb
```diff
@@ -7,7 +7,7 @@ class Railtie < Rails::Railtie
                              'X-Permitted-Cross-Domain-Policies', 'X-Download-Options',
                              'X-Content-Type-Options', 'Strict-Transport-Security',
                              'Content-Security-Policy', 'Content-Security-Policy-Report-Only',
-                             'Public-Key-Pins', 'Public-Key-Pins-Report-Only']
+                             'Public-Key-Pins', 'Public-Key-Pins-Report-Only, Referrer-Policy']
```
estark37: Is this supposed to be two separate strings instead of continuing the PKP-RO string?
```
'Public-Key-Pins-Report-Only', 'Referrer-Policy'
```

estark37:
Oh btw here's the link to the editor's draft that specs the separate header: https://w3c.github.io/webappsec-referrer-policy/#referrer-policy-header

lib/secure_headers/railtie.rb
oreoshake: Good :eyes:, I didn't catch that.

tomgilligan: Ahh! Soz

tomgilligan: Yep, addressing # 2 and typo in railtie now.

oreoshake:
Thanks @tomgilligan!

PR #252, opened 2016-05-08
Fix minor papercuts

stefansundin:
I just tested the gem and came across these minor papercuts.

- Fix minor papercuts in README.
- Add description to rake task to make it appear in `rake -T`.

Another thing I was wondering about was that it appears `frame_src` is deprecated (according to http://content-security-policy.com/), so does it make sense to remove it from the example in the README?

Thanks!


oreoshake:
Thanks! 🎊

PR #253, opened 2016-05-08
Fix false booleans being included in the CSP directive

stefansundin:
I was using `upgrade_insecure_requests: false` but I still got `upgrade-insecure-requests` in my header. It would appear this is a bug?


oreoshake:
Ha! What a simple mistake :smile:. The fix is great, but would you mind adding a test?

stefansundin:
Test added. Does it cover what you were looking for?

oreoshake:
Thanks! I'll cut a new gem shortly.

Issue #260, opened 2016-05-24
Firefox supports child-src in CSP


spdawson:
Testing with Firefox version 46.0.1, I am forced to use `frame-src` where I would like to use `child-src`, because `secure_headers` currently blocks `child-src` in the CSP header for Firefox; see https://github.com/twitter/secureheaders/blob/master/lib/secure_headers/headers/policy_management.rb#L99

However, the Firefox console logs this error:
```
Content Security Policy: Directive 'frame-src' has been deprecated. Please use directive 'child-src' instead.
```
which suggests that it might be time to permit `child-src` in the CSP header value for Firefox.

oreoshake:
:cool: I'll cut a PR later today unless someone beats me to it.

I've always wanted to be more intelligent around handling this cutover. i.e. we shouldn't have to specify a frame-src and child-src for cross-browser compatibility. Maybe this is a good reason to add some smarts and version-specific directive sets. 

* [ ] deprecate frame-src, encourage the use of child-src
* [ ] apply whichever directive the user agent understands.
  * [ ] figure out exactly when firefox started sending the deprecation warnings (46.0.1)
  * [ ] Add a TODO safari future cutover
  * [ ] For user agents with unknown support, send both?

oreoshake:
Ugh I took a stab at this but it became complicated when considering the precomputed, cached headers. The current code isn't well setup for version-specific policies. I'll have to step back and think about this instead of just slapping things together so they "work".

oreoshake:
Fixed in 3.4.0

PR #265, opened 2016-06-08
allows child-src values for firefox headers

oreoshake:
Partial fix for https://github.com/twitter/secureheaders/issues/260

It will not eliminate the deprecation warning, but it will at least allow the value.


oreoshake:
Fixed in https://github.com/twitter/secureheaders/commit/4e9da9dc6976b057137e2fe31a08c9818ab2d622

PR #267, opened 2016-06-10
wws: -> wss:

xtian:
I believe this is a typo.

### DIFF
--- a/README.md
+++ b/README.md
@@ -54,7 +54,7 @@ SecureHeaders::Configuration.default do |config|
     # directive values: these values will directly translate into source directives
     default_src: %w(https: 'self'),
     frame_src: %w('self' *.twimg.com itunes.apple.com),
-    connect_src: %w(wws:),
+    connect_src: %w(wss:),
     font_src: %w('self' data:),
     img_src: %w(mycdn.com data:),
     media_src: %w(utoob.com),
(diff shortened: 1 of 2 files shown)


jacobbednarz:
You are correct 👍🏼. This typo is used in a spec too! I wonder if it plays enough of a role that this typo should have caused a failure instead of being allowed. 

On mobile at the moment but will check the whole spec out later on.

jacobbednarz:
Looking further into it, the typo is not significant enough to cause a failure (at least the gem specs) since it is user input and there is no reason to try and validate it.

xtian:
Yep, I just thought it could potentially be confusing for people copy-pasting the README config.

jacobbednarz:
Agree ❤️

oreoshake:
>  This typo is used in a spec too

Yeah, it was copy/pasted from the readme at some point. @xtian mind updating it [here](https://github.com/twitter/secureheaders/blob/76fb8284e78e1233a7cdeaee205afd8690a74f4f/spec/lib/secure_headers/headers/policy_management_spec.rb#L26) too?

And yeah, I don't see much value in validating the content. It's moderately complex and error-prone which could lead to false positives. Also, if it's invalid, things will break anyways :smile:

xtian:
Did you mean to link to a different location? This PR already changes that line.

oreoshake:
Oops :coffee:

PR #272, opened 2016-07-13
hook include of helper to loading of ActionView

koenpunt:
## All PRs:

* [ ] Has tests
* [ ] Documentation updated

When using rspec with rails-controller-testing, specs would break due to incorrect load order. More info here: https://github.com/rails/rails-controller-testing/issues/24


oreoshake:
:cool:. Maybe wrap that in an block that checks if `ActiveSupport` is defined?

oreoshake:
Fixed in 3.4.0

PR #278, opened 2016-07-28
Add missing commas in the example configuration

vanderhoorn:


oreoshake:
Thanks. The CI failures are obviously unrelated (it looks like travis is ignoring the `bundler_args` value and trying to install guard)

PR #279, opened 2016-08-01
Minor change in domain

lalithr95:
This PR updates a minor change in documentation of domain used in example.

## All PRs:

* [ ] Has tests
* [x] Documentation updated

## Adding a new header

* What does it do?

## Adding a new CSP directive

* What does it do?


jacobbednarz:
LGTM

Issue #275, opened 2016-07-20
Add support for Permissions-Policy header


oreoshake:
Note: this issue used to focus on feature policy. Feature policy has been replaced with permissions-policy. While the two aren't equal, they are somewhat interchangeable for the discussion to this point.

Add support for [Feature Policy](https://wicg.github.io/feature-policy/). Feature policy needs to have support for highly dynamic policies just like CSP.

See [Mike West's AppSec EU presentation @39:21](https://youtu.be/0qu9oFaHuY0?t=2361)

> - Is the header supported by any user agent? If so, which?

A prototype is in chrome (ship in Q4?)

> - What does it do?

Enables/disables features in web browsers/

> - What are the valid values for the header?

Examples:

```
Feature-Policy: {"disable":["geolocation"], "target":["https://example.com"]}

Feature-Policy: {"disable":["webrtc"], "target":["https://example.com"]},
        {"enable":["webrtc"], "target":["https://other.com"]}
```

> - Where does the specification live?

https://wicg.github.io/feature-policy/

jacobbednarz:
Looks great! Definitely keen to get on board with this one (even as a guinea pig just for chrome).

oreoshake:
Abstracting out the dynamic part of how CSP is handled will help with this and https://github.com/twitter/secureheaders/issues/256 (setting two policies)

oreoshake:
This is in Chrome behind a flag now, might be worth adding support.

jacobbednarz:
This is getting some traction and I'm wondering if we want to take a pass at getting this into the library to at least be ready for when more browsers ship this.

oreoshake:
We should probably file an issue in rails/rails too.  I'm not sure if the syntax/values have changed and I'm too swamped to write out a more well thought out issue. We should mention that it'll need a similar API to the CSP support (dynamic policies)

jacobbednarz:
sounds like a plan - i might do it in the coming days seeing how we are on rails 5 now too.

jacobbednarz:
FWIW, I've raised this with Rails at https://github.com/rails/rails/pull/33439 🤞 it gets in a one less reason to have this gem maintained.

oreoshake:
Now that https://github.com/rails/rails/pull/33439 has shipped this feature ahead of secure_headers, maybe it's time to officially archive this project 😄 🚀

reedloden:
[...]

PR #281, opened 2016-08-09
Setting two CSP headers, abstract out dynamic pieces

### LINKED ISSUE #256: Set two csp headers (enforced and report only)
Is there a way using secure headers, to set both an enforced and report only header? A lot of people might want to do this in order to experiment with a different policy configuration and see what violation may occur before actually enforcing it.

oreoshake:
Fixes #256, #281
* [x] `@modified` bit strategy isn't working perfectly, I'm seeing objects being created when policies are not modified.
* [x] Has tests
* [x] Documentation updated

Setting two headers has been an often requested feature that has always had a workaround but was never a first class feature. This PR allows setting both headers.

In the process of supporting both, I decided to move away from hash-backed configs and towards class-based configs - mostly because it made debugging things very hard and mistyping a hash key is far to easy. This has the added benefit of not having to perform "deep clones" all the time and relying on hash voodoo. This is much clearer even if some of the operations are just glorified hash operations. 

This also adds a `@modified` attribute which signals whether the cached headers can be used or not. This removes the need to track a `dynamic_csp`, choose between two configs, and use the `idempotent_operations` call which was all just a bit too funky and created the need for a lot of boilerplate code.

When writing support for `override_content_security_policy_directives/append_content_security_policy_directives`, I realized that the default policy is not very user-friendly and insecure at the same time. Updating the default policy should make it easier for people to use the default setting and make their policy dynamic as needed. The new policy is less restrictive for the most part, with the exception of `object-src 'none'` which is just an unsafe default.

The reason the `DynamicConfig` stuff is extracted into a module with some hackery is so that it can be reused when implementing [FeaturePolicy](https://github.com/twitter/secureheaders/issues/275) which will benefit from the dynamic helper functions and non-hash-backed data. I like the idea, but the implementation looks a bit janky. I'll look into how ActiveRecord handles it's `changed?` function. I'm also questioning the use of `update_directive`/`directive_value`.

### DIFF
--- a/lib/secure_headers.rb
+++ b/lib/secure_headers.rb
@@ -106,11 +106,11 @@ module SecureHeaders
     def append_content_security_policy_directives(request, additions, target = nil)
       config, target = config_and_target(request, target)
 
-      if [:both, :enforced].include?(target) && config.csp != OPT_OUT
+      if [:both, :enforced].include?(target) && !config.csp.opt_out?
         config.csp.append(additions)
       end
 
-      if [:both, :report_only].include?(target) && config.csp_report_only != OPT_OUT
+      if [:both, :report_only].include?(target) && !config.csp_report_only.opt_out?
         config.csp_report_only.append(additions)
       end
 
--- a/lib/secure_headers.rb
+++ b/lib/secure_headers.rb
@@ -34,6 +34,10 @@ module SecureHeaders
       self.class.instance
     end
 
+    def opt_out?
+      true
+    end
+
     alias_method :[], :boom
     alias_method :[]=, :boom
     alias_method :keys, :boom
@@ -75,7 +79,7 @@ module SecureHeaders
       config, target = config_and_target(request, target)
 
       if [:both, :enforced].include?(target)
-        if config.csp == OPT_OUT
+        if config.csp.opt_out?
           config.csp = ContentSecurityPolicyConfig.new({})
         end
 
@@ -83,7 +87,7 @@ module SecureHeaders
       end
 
       if [:both, :report_only].include?(target)
-        if config.csp_report_only == OPT_OUT
+        if config.csp_report_only.opt_out?
           config.csp_report_only = ContentSecurityPolicyReportOnlyConfig.new({})
         end
 
--- a/lib/secure_headers.rb
+++ b/lib/secure_headers.rb
@@ -155,11 +155,11 @@ module SecureHeaders
       config = config_for(request, prevent_dup = true)
       headers = config.cached_headers
 
-      if config.csp != OPT_OUT && config.csp.modified?
+      if !config.csp.opt_out? && config.csp.modified?
         headers = update_cached_csp(config, headers, request.user_agent, CSP::CONFIG_KEY)
       end
 
-      if config.csp_report_only != OPT_OUT && config.csp_report_only.modified?
+      if !config.csp_report_only.opt_out? && config.csp_report_only.modified?
         headers = update_cached_csp(config, headers, request.user_agent, CSP::REPORT_ONLY_CONFIG_KEY)
       end
 
@@ -234,9 +234,9 @@ module SecureHeaders
     def guess_target(config)
       if !config.csp.opt_out? && !config.csp_report_only.opt_out?
         :both
-      elsif config.csp != OPT_OUT
+      elsif !config.csp.opt_out?
         :enforced
-      elsif config.csp_report_only != OPT_OUT
+      elsif !config.csp_report_only.opt_out?
         :report_only
       else
         :both
@@ -310,24 +310,6 @@ module SecureHeaders
     # Returns a CSP [header, value] array
     def csp_header_for_ua(headers, request)
       headers[CSP.ua_to_variation(UserAgent.parse(request.user_agent))]
-    end
-
-    # Private: optionally build a header with a given configure
-    #
-    # klass - corresponding Class for a given header
-    # config - A string, symbol, or hash config for the header
-    # user_agent - A string representing the UA  (only used for CSP feature sniffing)
-    #
-    # Returns a 2 element array [header_name, header_value] or nil if config
-    # is OPT_OUT
-    def make_header(klass, header_config, user_agent = nil)
-      unless header_config == OPT_OUT
-        if klass == CSP
-          klass.make_header(header_config, user_agent)
-        else
-          klass.make_header(header_config)
-        end
-      end
     end
   end
 
--- a/lib/secure_headers.rb
+++ b/lib/secure_headers.rb
@@ -66,9 +66,19 @@ module SecureHeaders
     #
     # additions - a hash containing directives. e.g.
     #    script_src: %w(another-host.com)
-    def append_content_security_policy_directives(request, additions)
+    def append_content_security_policy_directives(request, additions, target=:both)
+      unless [:both, :enforced, :report_only].include?(target)
+        raise "Unrecognized target: #{target}. Must be [:both, :enforced, :report_only]"
+      end
+
       config = config_for(request)
-      config.dynamic_csp = CSP.combine_policies(config.current_csp, additions)
+      if [:both, :enforced].include?(target) && config.current_csp != OPT_OUT
+        config.dynamic_csp = CSP.combine_policies(config.current_csp, additions)
+      end
+      if [:both, :report_only].include?(target) && config.current_csp_report_only != OPT_OUT
+        config.dynamic_csp_report_only = CSP.combine_policies(config.current_csp_report_only, additions)
+      end
+
       override_secure_headers_request_config(request, config)
     end
 
@@ -109,7 +119,11 @@ module SecureHeaders
     def header_hash_for(request)
       config = config_for(request)
       unless ContentSecurityPolicy.idempotent_additions?(config.csp, config.current_csp)
-        config.rebuild_csp_header_cache!(request.user_agent)
+        config.rebuild_csp_header_cache!(request.user_agent, CSP::CONFIG_KEY)
+      end
+
+      unless ContentSecurityPolicy.idempotent_additions?(config.csp_report_only, config.current_csp_report_only)
+        config.rebuild_csp_header_cache!(request.user_agent, CSP::REPORT_ONLY_CONFIG_KEY)
       end
 
       use_cached_headers(config.cached_headers, request)
@@ -190,7 +204,9 @@ module SecureHeaders
         ALL_HEADER_CLASSES
       else
         HTTP_HEADER_CLASSES
-      end
+      end.map do |klass|
+        klass::CONFIG_KEY
+      end.concat([CSP::REPORT_ONLY_CONFIG_KEY])
     end
 
     # Private: takes a precomputed hash of headers and returns the Headers
@@ -198,9 +214,9 @@ module SecureHeaders
     #
     # Returns a hash of header names / values valid for a given request.
     def use_cached_headers(headers, request)
-      header_classes_for(request).each_with_object({}) do |klass, hash|
-        if header = headers[klass::CONFIG_KEY]
-          header_name, value = if klass == CSP
+      header_classes_for(request).each_with_object({}) do |config_key, hash|
+        if header = headers[config_key]
+          header_name, value = if [CSP::CONFIG_KEY, CSP::REPORT_ONLY_CONFIG_KEY].include?(config_key)
             csp_header_for_ua(header, request)
           else
             header
--- a/lib/secure_headers.rb
+++ b/lib/secure_headers.rb
@@ -127,16 +127,19 @@ module SecureHeaders
     # in Rack middleware.
     def header_hash_for(request)
       config = config_for(request)
-      puts "final config#{ config.inspect}"
+      puts "\nfinal config\n#{ config.inspect}"
       unless ContentSecurityPolicy.idempotent_additions?(config.csp, config.current_csp)
         config.rebuild_csp_header_cache!(request.user_agent, CSP::CONFIG_KEY)
       end
+
+      puts "\nafter enforced changes\n#{ config.inspect}"
 
       unless ContentSecurityPolicy.idempotent_additions?(config.csp_report_only, config.current_csp_report_only)
         config.rebuild_csp_header_cache!(request.user_agent, CSP::REPORT_ONLY_CONFIG_KEY)
       end
 
-      puts "final config with cache#{ config.inspect}"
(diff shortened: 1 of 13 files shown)


jacobbednarz:
Where would you put this in terms of completion? Would you be confident in trying this out on an existing implementation yet? (with the exception of the `@modified` issue you mentioned above)

oreoshake:
I wouldn't use this implementation yet. The `@modified` bug may uncover other things. I'll have time to finish this out this week though.

jacobbednarz:
All good 👍  Just checking whether I should run this against our internal test suite and see if it pops up with anything but I'll hold off until the `@modified` stuff is tackled.

lib/secure_headers/configuration.rb
```diff
@@ -121,7 +119,8 @@ def deep_copy_if_hash(value)
     def initialize(&block)
       self.hpkp = OPT_OUT
       self.referrer_policy = OPT_OUT
-      self.csp = self.class.send(:deep_copy, CSP::DEFAULT_CONFIG)
+      self.csp = ContentSecurityPolicyConfig::DEFAULT
```
jacobbednarz: Would smarter defaults be to swap these conditions and allow a report only by default?

```
self.csp = OPT_OUT
self.csp_report_only = ContentSecurityPolicyConfig::DEFAULT
```

Or would this break backwards compatibility with existing implementations?

lib/secure_headers/configuration.rb
```diff
@@ -196,16 +191,42 @@ def secure_cookies=(secure_cookies)
...
-      if self.dynamic_csp
-        raise IllegalPolicyModificationError, "You are attempting to modify CSP settings directly. Use dynamic_csp= instead."
+      if new_csp == OPT_OUT
+        @csp = OPT_OUT
+      else
+        if new_csp.is_a?(ContentSecurityPolicyConfig)
+          @csp = new_csp.dup
+        else
+          if new_csp[:report_only]
+            Kernel.warn "#{Kernel.caller.first}: [DEPRECATION] `#csp=` was supplied a config with report_only: true. Use #csp_report_only="
```
jacobbednarz: 👍

lib/secure_headers/configuration.rb
oreoshake: Well moving to report only shouldn't break existing implementations, but the previous default was an enforced policy. The change to enforce-by-default was made at the last major version bump iirc so there's no reason to go "backwards" here. This PR does change the default policy to something more realistic and secure at the same time.

jacobbednarz: Sounds perfectly sane :)

oreoshake:
I removed the WIP label. I still need to update the docs and take another glance for quality over the change, but this should be ready for use.

jacobbednarz:
Awesome! I'll pull this locally and run it through it's paces in our setup. :)

lib/secure_headers.rb
```diff
@@ -66,9 +103,17 @@ def override_content_security_policy_directives(request, additions)
     #
     # additions - a hash containing directives. e.g.
     #    script_src: %w(another-host.com)
-    def append_content_security_policy_directives(request, additions)
-      config = config_for(request)
-      config.dynamic_csp = CSP.combine_policies(config.current_csp, additions)
+    def append_content_security_policy_directives(request, additions, target = nil)
+      config, target = config_and_target(request, target)
+
+      if [:both, :enforced].include?(target) && config.csp != OPT_OUT
```
jacobbednarz: should this be using the `opt_out?` method? i.e.

```rb
if [:both, :enforced].include?(target) && !config.csp.opt_out?
```

lib/secure_headers.rb
```diff
@@ -150,18 +201,47 @@ def content_security_policy_style_nonce(request)
...
+
+    def config_and_target(request, target)
+      config = config_for(request)
+      target = guess_target(config) unless target
+      raise_on_unknown_target(target)
+      [config, target]
+    end
+
+    def guess_target(config)
+      if config.csp != OPT_OUT && config.csp_report_only != OPT_OUT
```
jacobbednarz: same question as https://github.com/twitter/secureheaders/pull/281/files#r75221331. does this need to be swapped to use `opt_out?` too?

lib/secure_headers.rb
oreoshake: Seems so! Fixed in 3afd1dd

lib/secure_headers.rb
```diff
@@ -190,17 +270,19 @@ def header_classes_for(request)
         ALL_HEADER_CLASSES
       else
         HTTP_HEADER_CLASSES
-      end
+      end.map do |klass|
```
oreoshake: Note to self: wat ⁉️

lib/secure_headers/headers/content_security_policy.rb
```diff
@@ -1,33 +1,43 @@
...
+      @config = if config.is_a?(Hash)
+        if config[:report_only]
+          ContentSecurityPolicyReportOnlyConfig.new(config || DEFAULT_CONFIG)
+        else
+          ContentSecurityPolicyConfig.new(config || DEFAULT_CONFIG)
+        end
+      elsif config.nil?
+        ContentSecurityPolicyConfig.new(DEFAULT_CONFIG)
+      else
+        config.dup
```
oreoshake: This `dup` is only necessary because we actually modify the underlying arrays of the config with lots of `!` methods. A naive attempt to remove the ! methods and the `dup` failed miserably. But this is a smell worth cleaning up.

lib/secure_headers.rb
jacobbednarz: I was about to comment on this and say a similar thing about the readability of this 😛  The `def blah; end.map do |klass|` definitely makes for it slightly difficult to follow. Is there an approach that we could take here to break this up a bit?

lib/secure_headers.rb
```diff
@@ -150,18 +201,47 @@ def content_security_policy_style_nonce(request)
...
+      config = config_for(request)
+      target = guess_target(config) unless target
+      raise_on_unknown_target(target)
+      [config, target]
+    end
+
+    def guess_target(config)
+      if !config.csp.opt_out? && !config.csp_report_only.opt_out?
+        :both
+      elsif config.csp != OPT_OUT
```
jacobbednarz: sorry to be a broken, nagging record but it looks like we've got a few instances of the `!= OPT_OUT` in here still. would you like me to comment on each one I find or are you OK to just do a giant find + replace?

lib/secure_headers.rb
oreoshake: This is going to take another non-trivial, but overdue refactoring of the way CSP vs CSPRO are handled. The result should be much cleaner. Passing around symbols and class objects with all sorts of special casing have trickled down.

lib/secure_headers.rb
oreoshake: I'll do a pass and ping you when this and the refactor mentioned below are ready for :eyes:

jacobbednarz:
Not sure if it's intended but running a previous report only policy on this branch now has a enforced policy by default. 

```
  config.csp = {
    report_only:  true,
    report_uri: %w(https://reporting-endpoint),
    # .. snip
  }
end
```

Results in the `Content-Security-Policy` from https://github.com/twitter/secureheaders/pull/281/files#diff-16a93318ee7771ce5483636b9959f4ecR95 and then my custom policy in `Content-Security-Policy-Report-Only`. This is fine if it's intended but we might want to explicitly communicate this if so.

oreoshake:
> Not sure if it's intended but running a previous report only policy on this branch now has a enforced policy by default.

That's definitely a breaking change. A missing test case for sure.

jacobbednarz:
wow, coveralls is super keen 😕

jacobbednarz:
Actually looks like we might have a regression here with the default policy being added here even when it's not defined. 

```
  config.csp_report_only = {
    report_uri: %w(https://reporting-endpoint),
    # .. snip
  }
end
```

Results in the same policies as https://github.com/twitter/secureheaders/pull/281#issuecomment-240576590

oreoshake:
The case in https://github.com/twitter/secureheaders/pull/281#issuecomment-240590109 is a little tricky. Opting out of the default header is explicit extra step. Documentation can help with that and I'm not sure heuristically determining whether only the RO header should be set is a good idea.

jacobbednarz:
Ah, makes sense. So if one was looking to _only_ use a report only policy, it should be something like this?

```
  config.csp = SecureHeaders::OPT_OUT
  config.csp_report_only = {
    report_uri: %w(https://reporting-endpoint),
    # .. snip
  }
end
```

oreoshake:
Correct. It doesn't feel right, however. Maybe adding smarts around detecting intention won't be so bad.

jacobbednarz:
> Correct. It doesn't feel right, however.

Seems like a reasonable pattern to follow seeing that is how all the other directives are removed/ignored. Perhaps a sensible default of `SecureHeaders::OPT_OUT` is in order to ensure that if it's not explicitly set, you only get a read only policy?

Using the following to have a report only policy results in `wrong number of arguments (2 for 0..1) (ArgumentError)`

```rb
  config.csp = SecureHeaders::OPT_OUT
  config.csp_report_only = {
    report_uri: %w(https://reporting-endpoint),
    # .. snip
  }
end
```

oreoshake:
> Perhaps a sensible default of SecureHeaders::OPT_OUT is in order to ensure that if it's not explicitly set, you only get a read only policy?

Are you saying to make `csp`/`csp_report_only` default to `OPT_OUT` or are you saying something like:

>if you set `csp_report_only` and `csp == DEFAULT` at that time, then opt out of `csp` and set `csp_report_only`. 

The latter option was what I was going to try.

oreoshake:
> Using the following to have a report only policy results in wrong number of arguments (2 for 0..1) (ArgumentError)
> 
> ```
> SecureHeaders::Configuration.default do |config|
>   config.csp = SecureHeaders::OPT_OUT
>   config.csp_report_only = {
>     report_uri: %w(https://reporting-endpoint),
>     # .. snip
>   }
> end
> ```

@jacobbednarz can you provide the full stacktrace/config? I just plopped that in and it worked (but needed a `default-src` to be set). Certainly no `ArgumentError`s for me.

jacobbednarz:
Gemfile (running at 7600b90)
```
gem 'secure_headers', git: 'https://github.com/twitter/secureheaders.git',      
                      ref: 'dynamic-csp-config-abstraction'
```

Config (some of these directive values have been changed but it still results in the same error).
```rb
  config.csp = SecureHeaders::OPT_OUT
  config.csp_report_only = {
    report_uri: %w(https://my-report-only-uri),
    preserve_schemes: true,

    default_src: %w(* data:),
    img_src: %w(* data:),
    style_src: %w(* 'self' 'unsafe-inline'),
    font_src: %w('self' data:),
    script_src: %w(
      *
      data:
      'self'
      'unsafe-eval'
      'unsafe-inline'
    ),
    connect_src: %w(*),
    base_uri: %w(*),
  }
end
```

Stacktrace
```
/opt/boxen/rbenv/versions/2.2.5/lib/ruby/gems/2.2.0/bundler/gems/secureheaders-7600b900c666/lib/secure_headers.rb:25:in `boom': wrong number of arguments (2 for 0..1) (ArgumentError)
	from /Users/Jacob/src/marketplace/config/initializers/http_security_headers.rb:40:in `block in <top (required)>'
	from /opt/boxen/rbenv/versions/2.2.5/lib/ruby/gems/2.2.0/bundler/gems/secureheaders-7600b900c666/lib/secure_headers/configuration.rb:34:in `instance_eval'
	from /opt/boxen/rbenv/versions/2.2.5/lib/ruby/gems/2.2.0/bundler/gems/secureheaders-7600b900c666/lib/secure_headers/configuration.rb:34:in `override'
	from /Users/Jacob/src/marketplace/config/initializers/http_security_headers.rb:39:in `<top (required)>'
	from /opt/boxen/rbenv/versions/2.2.5/lib/ruby/gems/2.2.0/gems/activesupport-3.2.22.4/lib/active_support/dependencies.rb:245:in `load'
	from /opt/boxen/rbenv/versions/2.2.5/lib/ruby/gems/2.2.0/gems/activesupport-3.2.22.4/lib/active_support/dependencies.rb:245:in `block in load'
	from /opt/boxen/rbenv/versions/2.2.5/lib/ruby/gems/2.2.0/gems/activesupport-3.2.22.4/lib/active_support/dependencies.rb:236:in `load_dependency'
	from /opt/boxen/rbenv/versions/2.2.5/lib/ruby/gems/2.2.0/gems/activesupport-3.2.22.4/lib/active_support/dependencies.rb:245:in `load'
	from /opt/boxen/rbenv/versions/2.2.5/lib/ruby/gems/2.2.0/gems/railties-3.2.22.4/lib/rails/engine.rb:593:in `block (2 levels) in <class:Engine>'
	from /opt/boxen/rbenv/versions/2.2.5/lib/ruby/gems/2.2.0/gems/railties-3.2.22.4/lib/rails/engine.rb:592:in `each'

jacobbednarz:
I agree with 

> if you set csp_report_only and csp == DEFAULT at that time, then opt out of csp and set csp_report_only.

as it makes the most sense to me.

oreoshake:
:cool: added that in 3225ed6 and I believe I fixed your error too. I don't have an explanation, however.

jacobbednarz:
My apologies, it was never the default `csp`/`csp_report_only` causing the issues - it is actually one of our overrides that was doing it. I still had `config.csp` in one of our overrides despite having it set to `SecureHeaders::OPT_OUT` 😞

jacobbednarz:
This is now passing our internal test suite too! Very nice effort ✨

oreoshake:
@jacobbednarz are you using this now? I don't have any plans to use this feature at this time and I'm a little uncomfortable deploying a feature I'm not using. I'd love it if this code could bake in a real world scenario for some time.

jacobbednarz:
I'm happy to get our app running this branch however we are currently only using a report only policy but there are plans in the near future to look at an enforced policy for some of our directives so if you don't mind holding this pull request open for a bit,  I'd be happy to get this running in our production environment.

oreoshake:
:cool: I'm in no rush at this time. I'm not using it :smile:.

If anyone else has a chance to test drive it, that'd be :sparkles:

jacobbednarz:
This has been working great in production without any regressions identified using a `report-only` policy. We are still a couple of weeks off an enforced policy but I will keep you posted.

oreoshake:
I'm going to merge this branch and do a pre-release

oreoshake:
Released in 3.5.0.pre

PR #283, opened 2016-08-18
add support for named appends

### LINKED ISSUE #250: Support named appends
Similar to named overrides, this would allow me to have one file that tracks everything CSP and then just use named appends when I need to supplement the existing policy with additional things.

oreoshake:
Fixes #250
## All PRs:

* [x] Has tests
* [x] Documentation updated

@reedloden was this what you had in mind?

This allows you to define blocks of code that produce a hash of additions to be added to the CSP for a given request. The current request object is passed to the block. The result of the hash is passed to `append_content_security_policy_directives`.

The use case _usually_ is related to loosening a policy's restrictions. While we do provide `SecureHeaders::Configuration.override` blocks, something like this may be useful with `override_content_security_policy_directives` as well. I think that the demand for override support like this is pretty rare so I'll wait until there's a need. One example I can think of would be needing to override a directive with `'none'`, but chances are anything that needs to be set to `none` would be set in the default config.

Example use:

```ruby
def show
  if include_widget?
    @widget = widget.render
    use_content_security_policy_named_append(:widget_partial)
  end
end


SecureHeaders::Configuration.named_append(:widget_partial) do |request|
  if request.controller_instance.current_user.in_test_bucket?
    { child_src: %w(beta.thirdpartyhost.com) }
  else
    { child_src: %w(thirdpartyhost.com) }
  end
end
```

Blocked by #281 - This feature needs to be aware of the `csp`/`csp_report_only` config separation and needs to support the `target` attribute.

### DIFF
--- a/README.md
+++ b/README.md
@@ -94,6 +94,61 @@ use SecureHeaders::Middleware
 
 All headers except for PublicKeyPins have a default value. See the [corresponding classes for their defaults](https://github.com/twitter/secureheaders/tree/master/lib/secure_headers/headers).
 
+## Named Appends
+
+Named Appends are blocks of code that can be reused and composed during requests. e.g. If a certain partial is rendered conditionally, and the csp needs to be adjusted for that partial, you can create a named append for that situation. The value returned by the block will be passed into `append_content_security_policy_directives`. The current request object is passed as an argument to the block for even more flexibility.
+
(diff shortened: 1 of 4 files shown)


oreoshake:
I disabled @coveralls comments for now.

PR #287, opened 2016-09-06
handle the case where nonces weren't populated for default csp config…

### LINKED ISSUE #285: Nonce is not added to CSP header in rails 5
## Steps to reproduce

1. On a brand new rails 5.0.0.1 project add the `secure_headers` gem.
2. Create an initializer including `SecureHeaders::Configuration.default` 
3. Create an action and view including:

```
<h1>Hello</h1>

<%= nonced_javascript_tag do %>
  alert("world!");
<% end %>
```

### Expected outcome

The script tag is rendered including a `nonce` attribute and a `nonce-X` origin is added to the `Content-Security-Policy` header.

### Actual outcome

The script tag correctly includes a nonce attribute, but there's no nonce origin in the `Content-Security-Policy` HTTP header, so the browser refuses to execute the inline script.

oreoshake:
Fixes #285

Because nonces are treated a meta config values they were not included when being copied into undefined directive values.

e.g.

```ruby
  config.csp = {
    default_src: %w(https:),
  }
end
```

```erb
<h1>Hello</h1>

<%= nonced_javascript_tag do %>
  alert("world!");
<% end %>
```

Would result in the nonce getting dropped from the `script-src` value that should have been copied from `default-src`.

/cc @afcapel if you have time to test this

### DIFF
--- a/lib/secure_headers/headers/policy_management.rb
+++ b/lib/secure_headers/headers/policy_management.rb
@@ -53,16 +53,6 @@ module SecureHeaders
     FRAME_ANCESTORS = :frame_ancestors
     PLUGIN_TYPES = :plugin_types
 
-    # These are directives that do not inherit the default-src value. This is
-    # useful when calling #combine_policies.
-    NON_FETCH_SOURCES = [
-      BASE_URI,
-      FORM_ACTION,
-      FRAME_ANCESTORS,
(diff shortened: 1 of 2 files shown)


afcapel:
@oreoshake this solves the issue for me :) 

Thanks for the prompt fix and the overall awesome project!

PR #288, opened 2016-09-10
Fixing README documentation on csp_report_only

btsai:
## All PRs:

* [x] Documentation updated

`csp_report_only` is available only in 3.5.0_pre and onwards, but if you run bundle install with `gem 'secure_headers`` in your Gemfile, it will pull 3.4.1 which doesn't have `csp_report_only` available.


oreoshake:
Good call. I'll merge it when I'm in front of a computer.

PR #290, opened 2016-09-29
Fix readme

lalithr95:
## All PRs:

Remove irrelevant`If **` in Readme.

* [ ] Has tests
* [x] Documentation updated

## Adding a new header

* What does it do?

## Adding a new CSP directive

* What does it do?

### DIFF
--- a/README.md
+++ b/README.md
@@ -122,8 +122,6 @@ Configuration.default do |config|
 end
 ```
 
-If **
-
 ## Named Appends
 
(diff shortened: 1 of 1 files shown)


lalithr95:
Not sure why build failed, I don't see rebuild option. Can any one click on rebuild ?

oreoshake:
The build errors should go away with https://github.com/twitter/secureheaders/issues/291

PR #291, opened 2016-09-29
restrict term-ansicolor so it works on 1.9

oreoshake:
This might fix the issues @lalithr95 is running into in https://github.com/twitter/secureheaders/pull/290


lalithr95:
Thanks ! @oreoshake

PR #295, opened 2016-10-04
Attempt at breaking up the documentation

### LINKED ISSUE #274: Split up monolithic documentation
The README is too long and hard to follow. Instead of trimming content, I think it would make sense to split it up into a `docs` directory with discrete files (creating a config, dynamic policies, using different frameworks, cookie management, etc). 

This was originally discussed in https://github.com/twitter/secureheaders/issues/269#issuecomment-228167780 but I'm creating a specific issue for this purpose in case anyone sees this and wants to take on a fairly beginner-friendly task.

The docs could probably use some cleaning up as well as more elaboration on dense topics. 

/cc @connorshea as the originator of the idea.

rcaught:
Fixes #274.


docs/defaults.md
```diff
@@ -0,0 +1,26 @@
+## Default values
```
jacobbednarz: This might be better suited in the `configuration.md` so people have a quick and easy point of reference for starting out?

oreoshake: I agree. This page, docs/README.md, and docs/configuration.md should probably all live in the original readme

rcaught:
@oreoshake, how's this looking now?

oreoshake:
@rcaught I think the "main" readme is still too thin on content. Please move `docs/configuration.md` back into the main readme.

oreoshake:
Hey @rcaught, will have a chance to update this sometime soon? If not, I'm happy to take it from here.

PR #296, opened 2016-10-12
Set secure cookies on interleaved http/https calls correctly

anujdas:
## All PRs:

* [ ] Has tests
* [ ] Documentation updated

## Adding a new header

* What does it do?

## Adding a new CSP directive

* What does it do?

The current HTTP/HTTPS detection code mutates the cookie config in place
in order to switch between secure and non-secure cookies. However, since
the original global secure_headers config was passed around everywhere,
the very first HTTP request received would modify the global cookie
config, resulting in :secure => false for all subsequent requests
regardless of SSL. The solution is to dup the cookie config before using
it, which is a pattern followed for every other configuration type
anyways; a better long-term fix would be using non-mutating methods
instead.


oreoshake:
Good catch. I'll wait for CI and push a new gem.

oreoshake:
released in 3.5.0

anujdas:
Thanks @oreoshake for the quick release!

PR #297, opened 2016-10-25
Add missing Referrer Policies

jeremydurham:
## All PRs:

* [x] Has tests
* [ ] Documentation updated

## Adding a new header

* What does it do?

## Adding a new CSP directive

* What does it do?

  * Add same-origin, strict-origin and strict-origin-when-cross-origin


oreoshake:
Thanks!

Thinking ahead, we might want to add smarts around what is supported since fallback behavior is inconsistent. One browser would strip referrers (fails closed) in all cases on unknown tokens and the other sends referrers in all cases (fails open).

Issue #299, opened 2016-11-02
NoMethodError upon initialize


jborrey:
### Expected outcome

`SecureHeaders::Configuration.default` should not crash.

### Actual outcome

`SecureHeaders::Configuration.default` crashes.

In a ruby terminal
```ruby
require 'secure_headers'
SecureHeaders::Configuration.default
```

```
NoMethodError: undefined method `>=' for nil:NilClass
lib/secure_headers/headers/content_security_policy.rb:217:in `supported_directives'
```

Stacktrace:

```
secure_headers-3.5.0/lib/secure_headers/headers/content_security_policy.rb:217:in `supported_directives'
secure_headers-3.5.0/lib/secure_headers/headers/content_security_policy.rb:65:in `normalize_child_frame_src'
secure_headers-3.5.0/lib/secure_headers/headers/content_security_policy.rb:31:in `initialize'
secure_headers-3.5.0/lib/secure_headers/headers/policy_management.rb:192:in `new'
secure_headers-3.5.0/lib/secure_headers/headers/policy_management.rb:192:in `make_header'
secure_headers-3.5.0/lib/secure_headers/configuration.rb:297:in `block in generate_csp_headers_for_config'
secure_headers-3.5.0/lib/secure_headers/configuration.rb:296:in `each'
secure_headers-3.5.0/lib/secure_headers/configuration.rb:296:in `generate_csp_headers_for_config'
secure_headers-3.5.0/lib/secure_headers/configuration.rb:289:in `generate_csp_headers'
secure_headers-3.5.0/lib/secure_headers/configuration.rb:276:in `cache_headers!'
secure_headers-3.5.0/lib/secure_headers/configuration.rb:74:in `add_configuration'
secure_headers-3.5.0/lib/secure_headers/configuration.rb:18:in `default'
```

Am I missing something in configuration?

oreoshake:
@jborrey Wow. So fail, but I can't reproduce. And there's lots of tests using the same code: https://github.com/twitter/secureheaders/search?utf8=%E2%9C%93&q=Configuration.default

What version of ruby are you using? 

```
[Neils-MacBook-Pro secureheaders (master)]$ gem install secure_headers -v 3.5.0
Fetching: secure_headers-3.5.0.gem (100%)
Successfully installed secure_headers-3.5.0
Parsing documentation for secure_headers-3.5.0
Installing ri documentation for secure_headers-3.5.0
Done installing documentation for secure_headers after 0 seconds
1 gem installed
[...]

PR #300, opened 2016-11-03
useragent gem version pinning

jborrey:
If the `useragent` version is too low, parsing user agents based on a
string like `"Chrome"` will not populate the version which is required
for other logic. Pinning the version will prevent random crashes from
not having the right version.

### DIFF
--- a/secure_headers.gemspec
+++ b/secure_headers.gemspec
@@ -15,5 +15,5 @@ Gem::Specification.new do |gem|
   gem.test_files    = gem.files.grep(%r{^(test|spec|features)/})
   gem.require_paths = ["lib"]
   gem.add_development_dependency "rake"
-  gem.add_dependency "useragent"
+  gem.add_dependency "useragent", "0.16.8"
 end


secure_headers.gemspec
```diff
@@ -15,5 +15,5 @@ Gem::Specification.new do |gem|
   gem.test_files    = gem.files.grep(%r{^(test|spec|features)/})
   gem.require_paths = ["lib"]
   gem.add_development_dependency "rake"
-  gem.add_dependency "useragent"
+  gem.add_dependency "useragent", "0.16.8"
```
reedloden: Not a fan of pinning to a specific version, as it makes it a big pain to fix when there's a security issue or something with a gem. Can't we just do `"~> 0.16.8"` or something if that's the minimum version we actually need?

oreoshake:
As discussed [here](https://github.com/twitter/secureheaders/issues/299#issuecomment-258238019), I'm strongly against pinning `useragent`. The code should be resilient to `nil` with some unsurprising fallback behavior.

jborrey:
Closing in favor of [#301](https://github.com/twitter/secureheaders/pull/301) since this will take care of null user agents.

PR #301, opened 2016-11-07
handle null useragent version

stve:
## All PRs:

* [X] Has tests
* [X] Documentation updated (Not applicable)

This is a potential fix for #299, I'm not too familiar with useragent versioning but this seems like it will work. I've tested it against `0.10.0` which is mentioned in the issue.

I'm also not really loving `VERSION_NIL` as a constant name but couldn't think of anything better. 🤔

### DIFF
--- a/lib/secure_headers/headers/content_security_policy.rb
+++ b/lib/secure_headers/headers/content_security_policy.rb
@@ -9,6 +9,7 @@ module SecureHeaders
     # constants to be used for version-specific UA sniffing
     VERSION_46 = ::UserAgent::Version.new("46")
     VERSION_10 = ::UserAgent::Version.new("10")
+    FALLBACK_VERSION = ::UserAgent::Version.new("0")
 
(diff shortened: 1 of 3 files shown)


oreoshake:
I'm happy with the `VERSION_NIL` thing. Support always means that you're on at least version X, not at most.

Can you add a similar guard to https://github.com/twitter/secureheaders/blob/df0364633e6fd3b638dc3f5238b5a009204a96ba/lib/secure_headers/headers/policy_management.rb#L220 as well? It might be good guarding against all missing attributes, but we should handle all of the known bad attributes.

stve:
I add a check to `PolicyManagement` and also renamed the constant to `FALLBACK_VERSION` (I prefer that especially since `Version.new("0")` doesn't really define nil.

oreoshake:
🙇

jborrey:
Nice work guys!

PR #312, opened 2017-01-13
Update documentation: how to use gem with hanami

davydovanton:
I hope it will be helpful for other developers 🌟

### DIFF
--- a/README.md
+++ b/README.md
@@ -47,6 +47,24 @@ For Rails 2 or non-rails applications, an explicit statement is required to use 
 use SecureHeaders::Middleware
 ```
 
+### hanami
+
+If you want to use secure_headers with [hanami](http://hanamirb.org) applications you need to take two steps:
+
+First, you need to use `SecureHeaders::Middleware` middleware:
+```ruby
+# config.ru
+use SecureHeaders::Middleware
+```
+
+And second, you need to create `config/initializers/secure_headers.rb` for configurations:
+```ruby
+# config/initializers/secure_headers.rb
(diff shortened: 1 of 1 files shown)


jacobbednarz:
Hanami already has most of the secure headers [stuff built in](https://github.com/hanami/hanami/blob/master/lib/hanami/generators/app/application.rb.tt#L155) (and tweaked thanks to @oreoshake!).

What are you looking to get out of using Hanami and secure headers together that isn't already covered out of the box?

davydovanton:
Hey, you can use any middlewares with web interface outside hanami protection (for example sidekiq).

So, let me give you a real example (from my hanami application).

I use sidekiq middleware in my admin application (I mean [I mounted](https://github.com/davydovanton/ossboard/blob/master/apps/admin/config/routes.rb#L1) sidekiq middleware in admin routes). Admin application has this config line:
```ruby
security.x_frame_options 'DENY'
```

But when I try to request `/admin/sidekiq` path I have this option (`SAMEORIGIN`):
screenshot 2017-01-13 05 06 37

So, If I add secureheaders gem with `'DENY'` option I will get this result (`DENY `):
screenshot 2017-01-13 05 06 26

That's why I think that using secureheaders gem with hanami is right solution 👍

davydovanton:
@oreoshake could you look on this pr? :)

oreoshake:
Sorry for letting this sit around for so long. Please move this to a file under the docs directory.

oreoshake:
closing as stale

PR #322, opened 2017-04-02
Add support for `Expect-CT` HTTP header

### LINKED ISSUE #321: Support for Expect-CT header
This new header informs the browser that it is expected for your certificate to appear in a Certificate Transparency log.

Detailed explanation here: https://scotthelme.co.uk/a-new-security-header-expect-ct/
Browser support: proposed in Chrome, planned for Firefox (https://www.chromestatus.com/feature/5677171733430272)
Specification: https://datatracker.ietf.org/doc/draft-stark-expect-ct/

Sample header: `Expect-CT: enforce; max-age=30; report-uri https://scotthelme.report-uri.io/r/default/ct/enforce`

If `enforce` then the site will not be loaded if the certificate cant be found in a CT log. If absent, then it will only be reported.

jacobbednarz:
`Expect-CT` is a new security header that performs a request to a
certificate transparency log endpoint that will check whether the
presented certificate adhears to the UA's certificate transparency
policy.

Whilst building this out, I made the decision to use the hash
configuration syntax (similar to CSP) instead of the string based syntax
(like HSTS) to ensure we could validate and build various parts of the
value without needing to perform horrible regexes across it.

Fixes #321


jacobbednarz:
I've got this running in our CI suite at the moment and it's all 💚  here

lib/secure_headers.rb
```diff
@@ -51,6 +52,7 @@ def opt_out?
   CSP = ContentSecurityPolicy
 
   ALL_HEADER_CLASSES = [
+    ExpectCt,
```
oreoshake: I prefer to spell out the full name of the header spec.

lib/secure_headers/configuration.rb
```diff
@@ -116,7 +116,7 @@ def deep_copy_if_hash(value)
 
     attr_writer :hsts, :x_frame_options, :x_content_type_options,
       :x_xss_protection, :x_download_options, :x_permitted_cross_domain_policies,
-      :referrer_policy, :clear_site_data
+      :referrer_policy, :clear_site_data, :expect_ct
```
oreoshake: I prefer to spell out the full name of the header spec.

lib/secure_headers/headers/expect_ct.rb
```diff
@@ -0,0 +1,69 @@
+module SecureHeaders
+  class ExpectCtConfigError < StandardError; end
+
+  class ExpectCt
+    HEADER_NAME = 'Expect-CT'.freeze
+    CONFIG_KEY  = :expect_ct
+    INVALID_CONFIGURATION_ERROR = 'config must be a hash.'.freeze
```
oreoshake: I prefer `"`

lib/secure_headers/headers/expect_ct.rb
```diff
@@ -0,0 +1,69 @@
...
+        enforced_directive,
+        max_age_directive,
+        report_uri_directive
+      ].compact.join('; ').strip
+    end
+
+    def enforced_directive
+      # Unfortunately `if @enforced` isn't enough here in case someone
+      # passes in a random string so let's be specific with it to prevent
+      # accidental enforcement.
```
oreoshake: :+1:

lib/secure_headers/headers/expect_ct.rb
```diff
@@ -0,0 +1,69 @@
...
+      "enforce" if @enforced == true
+    end
+
+    def max_age_directive
+      "max-age=#{@max_age}" if @max_age
+    end
+
+    def report_uri_directive
+      "report-uri=\"#{@report_uri}\"" if @report_uri
```
oreoshake: `report-uri` is quoted. Interesting.

lib/secure_headers.rb
jacobbednarz: Good call! Fixed via 39cbb52

lib/secure_headers/headers/expect_ct.rb
jacobbednarz: This did strike me as odd too but, you know, specs? `¯\_(ツ)_/¯`

lib/secure_headers/headers/expect_ct.rb
jacobbednarz: Oh my! What caveman would ever commit files with mixed quotes?!?! 😛  Fixed in 9faf299

oreoshake:
whoops, can you add an assertion [here](https://github.com/twitter/secureheaders/blob/a85b80cca27106f357900f0ac930caec30f6be74/spec/spec_helper.rb#L36) that says this header is not applied by default.

jacobbednarz:
Of course! Added in c265600.

jacobbednarz:
We've been running this for over a week now and all is looking good on our front 👍

jacobbednarz:
This is pending release from one of the browser vendors to test the functionality confirms to the outlined spec. My comment above was merely, "this isn't blowing up our application with memory leaks or such".

jacobbednarz:
updates! https://groups.google.com/a/chromium.org/forum/m/#!msg/blink-dev/7_EgKuWAXjE/QJCzbBscBQAJ

jacobbednarz:
Looks like this is coming in Chrome 61
https://blog.chromium.org/2017/08/chrome-61-beta-javascript-modules.html<https://blog.chromium.org/2017/08/chrome-61-beta-javascript-modules.html?m=1>
________________________________

oreoshake:
Released in 3.7.0

PR #324, opened 2017-04-13
remove deprecation warning about frame_src

bobjflong:
`frame-src` was un-deprecated: https://w3c.github.io/webappsec-csp/#changes-from-level-2

Therefore I think this warning should be removed.

## All PRs:

* [ ✔️  ] Has tests
* [n/a?] Documentation updated

### DIFF
--- a/spec/lib/secure_headers/headers/content_security_policy_spec.rb
+++ b/spec/lib/secure_headers/headers/content_security_policy_spec.rb
@@ -86,8 +86,8 @@ module SecureHeaders
         expect(csp.value).to eq("default-src example.org")
       end
 
-      it "emits a warning when using frame-src" do
-        expect(Kernel).to receive(:warn).with(/:frame_src is deprecated, use :child_src instead./)
+      it "does not emit a warning when using frame-src" do
+        expect(Kernel).to_not receive(:warn)
         ContentSecurityPolicy.new(default_src: %w('self'), frame_src: %w('self')).value
       end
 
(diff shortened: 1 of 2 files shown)


spec/lib/secure_headers/headers/content_security_policy_spec.rb
```diff
@@ -86,8 +86,8 @@ module SecureHeaders
         expect(csp.value).to eq("default-src example.org")
       end
 
-      it "emits a warning when using frame-src" do
-        expect(Kernel).to receive(:warn).with(/:frame_src is deprecated, use :child_src instead./)
+      it "does not emit a warning when using frame-src" do
+        expect(Kernel).to_not receive(:warn).with(/:frame_src is deprecated, use :child_src instead./)
```
eugeneius: This is asserting that a _particular_ warning isn't logged; we probably want to assert that _no_ warnings are logged:

```ruby
expect(Kernel).to_not receive(:warn)
```

oreoshake:
Thanks @bobjflong! I agree with @eugeneius on the assertion.

Should we add a warning deprecating `child-src`? I'm a little foggy on the details but I think encouraging the use of `frame-src` and `worker-src` over `child-src` (which envelops frame/worker) is the direction the spec is going.

oreoshake:
> Should we add a warning deprecating child-src? 

☝️ wouldn't block shipping this, but this did prompt the conversation.

oreoshake:
There is no need to set both as the code prefers one, falls back to the other, and warns if the values do not match. Which setting it prefers should be swapped as well.

bobjflong:
Thanks, would it be possible to get a new version with these changes? Or maybe some advice on when a new version might be made available?

oreoshake:
@bobjflong release in 3.6.3

PR #326, opened 2017-04-26
remove `nonce_added?`

bobjflong:
`nonce_added?` was always returning truthy, so I think it can be erased.

It was either returning `true`, or the result of the `.each` block (which is `[:script_nonce, :style_nonce]`)


oreoshake:
I can't recall why that code was there in the first place, perhaps to prevent a nonce appearing twice in one directive? I would have hoped that a test would have failed with this change, exposing it's intent 😢  I'll take a closer look soon.

bobjflong:
Note that merging master added `nonce_added?` back to this branch.

oreoshake:
Fixed in https://github.com/twitter/secureheaders/pull/354. I finally understand the point of this method!

PR #329, opened 2017-05-10
Updated a few places where wrong wordage was being used for where and…

scouttyg:
… how to configure hashes

Mostly these were slight tweaks to the pages, along with one small tweak to the way hashes were being outputted to the error screen.


scouttyg:
Hey all, just checking back on this. Mostly tweaks, but think it could still help out people who might be confused.

oreoshake:
Whoops, this fell off my radar. Thanks @scouttyg

PR #330, opened 2017-05-25
Enable verbose mode in tests

stve:
I enabled warnings in one of my apps and saw that the secure_headers was the source of a number of warnings. This change enables warnings in rspec and fixes the existing warnings. I ran the test suite about 25-30 times to cover multiple seeds as possible, but there's a chance I still missed some.

## All PRs:

* [x] Has tests
* [x] Documentation updated

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -15,7 +15,7 @@ Please note that this project is released with a [Contributor Code of Conduct][c
 0. Configure and install the dependencies: `bundle install`
 0. Make sure the tests pass on your machine: `bundle exec rspec spec`
 0. Create a new branch: `git checkout -b my-branch-name`
-0. Make your change, add tests, and make sure the tests still pass
+0. Make your change, add tests, and make sure the tests still pass and that no warnings are raised
 0. Push to your fork and [submit a pull request][pr]
 0. Pat your self on the back and wait for your pull request to be reviewed and merged.
 
@@ -25,7 +25,7 @@ Here are a few things you can do that will increase the likelihood of your pull 
 - Keep your change as focused as possible. If there are multiple changes you would like to make that are not dependent upon each other, consider submitting them as separate pull requests.
 - Write a [good commit message](http://tbaggery.com/2008/04/19/a-note-about-git-commit-messages.html).
 
-## Releasing 
+## Releasing
 
 0. Ensure CI is green
 0. Pull the latest code
@@ -39,5 +39,3 @@ Here are a few things you can do that will increase the likelihood of your pull 
 
 - [How to Contribute to Open Source](https://opensource.guide/how-to-contribute/)
 - [Using Pull Requests](https://help.github.com/articles/about-pull-requests/)
-
-
--- a/lib/secure_headers.rb
+++ b/lib/secure_headers.rb
@@ -163,7 +163,7 @@ module SecureHeaders
     # returned is meant to be merged into the header value from `@app.call(env)`
     # in Rack middleware.
     def header_hash_for(request)
-      config = config_for(request, prevent_dup = true)
+      config = config_for(request, _prevent_dup = true)
       headers = config.cached_headers
       user_agent = UserAgent.parse(request.user_agent)
 
--- a/lib/secure_headers.rb
+++ b/lib/secure_headers.rb
@@ -163,7 +163,8 @@ module SecureHeaders
     # returned is meant to be merged into the header value from `@app.call(env)`
     # in Rack middleware.
     def header_hash_for(request)
-      config = config_for(request, _prevent_dup = true)
+      prevent_dup = true
+      config = config_for(request, prevent_dup)
       headers = config.cached_headers
       user_agent = UserAgent.parse(request.user_agent)
 
--- a/lib/secure_headers/configuration.rb
+++ b/lib/secure_headers/configuration.rb
@@ -31,7 +31,7 @@ module SecureHeaders
           raise NotYetConfiguredError, "#{base} policy not yet supplied"
         end
         override = @configurations[base].dup
-        override.instance_eval &block if block_given?
+        override.instance_eval(&block) if block_given?
         add_configuration(name, override)
       end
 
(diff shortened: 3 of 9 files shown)


lib/secure_headers.rb
```diff
@@ -163,7 +163,7 @@ def opt_out_of_all_protection(request)
     # returned is meant to be merged into the header value from `@app.call(env)`
     # in Rack middleware.
     def header_hash_for(request)
-      config = config_for(request, prevent_dup = true)
+      config = config_for(request, _prevent_dup = true)
```
oreoshake: This one seems odd. Is it saying use kwargs instead?

lib/secure_headers/configuration.rb
```diff
@@ -120,19 +120,36 @@ def deep_copy_if_hash(value)
...
     HASH_CONFIG_FILE = ENV["secure_headers_generated_hashes_file"] || "config/secure_headers_generated_hashes.yml"
-    if File.exists?(HASH_CONFIG_FILE)
+    if File.exist?(HASH_CONFIG_FILE)
       config = YAML.safe_load(File.open(HASH_CONFIG_FILE))
       @script_hashes = config["scripts"]
       @style_hashes = config["styles"]
     end
 
     def initialize(&block)
+      @cookies = nil
```
oreoshake: This is kind of a bummer. Same for https://github.com/twitter/secureheaders/pull/330/files#diff-16a93318ee7771ce5483636b9959f4ec. Should we just not use `@instance_var_name`?

lib/secure_headers/headers/policy_management.rb
```diff
@@ -348,9 +348,9 @@ def ensure_array_of_strings!(directive, source_expression)
       end
 
       def ensure_valid_sources!(directive, source_expression)
-        source_expression.each do |source_expression|
-          if ContentSecurityPolicy::DEPRECATED_SOURCE_VALUES.include?(source_expression)
-            raise ContentSecurityPolicyConfigError.new("#{directive} contains an invalid keyword source (#{source_expression}). This value must be single quoted.")
+        source_expression.each do |expression|
+          if ContentSecurityPolicy::DEPRECATED_SOURCE_VALUES.include?(expression)
+            raise ContentSecurityPolicyConfigError.new("#{directive} contains an invalid keyword source (#{expression}). This value must be single quoted.")
```
oreoshake: This one seems strange as well.

oreoshake:
Thanks as always @stve. Double bonus points for the process improvement.

Can you tell I don't enable warnings anywhere :smile: ?

lib/secure_headers.rb
stve: apparently `prevent_dup` wasn't used. I didn't want to remove it b/c I figured there was a reason it existed, but prefixing the arguement with an underscore suppresses the warning.

lib/secure_headers/configuration.rb
stve: yep, these are unfortunate. This was the most common warning: "warning: instance variable @ivar not initialized"

lib/secure_headers.rb
oreoshake: Strange. It's [definitely used](https://github.com/twitter/secureheaders/blob/fd559f2acc591c04e0c9563e3f9e857e7ce6e6c9/lib/secure_headers.rb#L233) and important.

As a side note: I'm ok with making strange changes to appease a tool in the name of keeping a clean list of warnings.

lib/secure_headers/headers/policy_management.rb
stve: the problem here was that the block parameter `source_expression` was identical to that of the outer parameter (`source_expression.each`)

lib/secure_headers.rb
stve: Ohhh. just realized, this is a method call, so I think it should actually just be `config = config_for(request, true)`

stve:
@oreoshake absolutely!

I'm a bit new to this as well, was actually inspired by the latest Bike Shed podcast episode with @tenderlove, @sgrif and @derekprior to avoid warnings in gems. 😎

lib/secure_headers.rb
oreoshake: 🤔 an early mentor of mine emphasized not passing around "magic booleans" (or any magic things really). I guess we could add a second method or treat the value as an options hash.

Maybe in 4.0 we can drop < 2.1 and use kwargs.

stve: I agree, kwargs would be a logical change here ✅  

To get around the warning, this would probably work too:

```ruby
prevent_dup = true
config = config_for(request, prevent_dup)
```

oreoshake:
> to avoid warnings in gems

A noble cause. Here I am, litterin' up logfiles and terminal screens unknowingly.

giphy

lib/secure_headers.rb
oreoshake: :+1:

CONTRIBUTING.md
```diff
@@ -15,7 +15,7 @@ Please note that this project is released with a [Contributor Code of Conduct][c
 0. Configure and install the dependencies: `bundle install`
 0. Make sure the tests pass on your machine: `bundle exec rspec spec`
 0. Create a new branch: `git checkout -b my-branch-name`
-0. Make your change, add tests, and make sure the tests still pass
+0. Make your change, add tests, and make sure the tests still pass and that no warnings are raised
```
oreoshake: Would this get enforced by CI? It looks like `--warnings` will still return a 0 exit code. It would be nice to enforce this.

stve: good point. I'm not sure if `--warnings` will, but maybe via `RUBYOPT` it can be enabled. I'll see if I can figure out a way to enforce it.

One thing to note is that there are a few deprecation warnings which we don't want to cause the suite to fail.

stve: This has me stumped. I can't seem to figure out how to make the suite fail when there are warnings outputted. 

I guess another route to go would be to add rubocop. I created a branch where I added rubocop and was able to bring it down to ~270 violations. Most of those remaining could be fixed either by overriding rubocop's default rules or via code changes (line length in particular is probably the most prominent violation).

oreoshake: > This has me stumped

Thanks for trying. I think we can :ship: what we have here if you think it's :+1: 

> I guess another route to go would be to add rubocop.

That would be great! I'm happy to help whittle this down.

sgrif: IIRC warnings go to STDERR, so you should be able to just enforce that STDERR is empty in CI

oreoshake: > IIRC warnings go to STDERR, so you should be able to just enforce that STDERR is empty in CI

:thinking: that might be a bit too strict. Other stderr might happen. I guess we can go with that strategy until we find it doesn't work. Any "other" stderr would not be the responsibility of the PR author, but hopefully adding this restriction doesn't discourage working a PR to completion.

stve: Locally (perhaps on Travis as well) i've been getting warnings from `pry` that would probably turn up in stderr as well.

stve:
@oreoshake if you think this is ok to merge as is, i'll open another PR to add rubocop to get that process started.

PR #332, opened 2017-05-31
Add RuboCop

stve:
This adds a basic RuboCop setup for us to work through. A few things to note:

1. I disabled the frozen string literal requirement entirely, only available in Ruby 2.3+
2. I disabled line/block/module length rules in `spec/`, these rules aren't really necessary in a test suite
3. `.ruboco_todo.yml` contains a large number of violations and their applicable files.

RuboCop's autocorrect feature fixes most of the violations, but results in a number of test suite failures so we'll need to work through them.

@oreoshake If you want to keep this PR open, I can start fixing the violations in this branch as well.


jacobbednarz:
💝  For tackling the violations, I've found ignoring a subset to not work well since they don't often get revisited or even known they were ignored in the first place. A more effective approach is passing through the changed files (per git) to rubocop as a part of CI. This ensures that when a file is next touched in any way the file is required to be updated to meet the new standards.

Happy to help out fixing violations too!

stve:
> A more effective approach is passing through the changed files (per git) to rubocop as a part of CI. This ensures that when a file is next touched in any way the file is required to be updated to meet the new standards.

@jacobbednarz absolutely! Ideally we can ditch as many exclusions as possible. This was my first time using `rubocop --auto-gen-config` to create `.rubocop_todo.yml` which is at least a nice way to step through everything. Perhaps it makes sense to remove all of the existing exclusions in `.rubocop.yml` and just add those to the list of violations to resolve. 

Do you have a default set of RuboCops that you configure/disable/etc. or do you prefer the default configuration?

> Happy to help out fixing violations too!

I can add you to my fork to work through together

oreoshake:
> Do you have a default set of RuboCops that you configure/disable/etc. or do you prefer the default configuration?

https://github.com/github/rubocop-github :question:

stve:
Interesting, I'd never seen rubocop-github before, looks like a sane set of rules, similar to some of the ones that I typically adopt on most of my projects as well. I'm game for using it. 

Presumably this is well maintained and used by GitHub?

oreoshake:
> Presumably this is well maintained and used by GitHub?

It's the base from which the github repo uses. There are some additional settings and local overrides but they're mostly custom cops or rails-specific. I'm very happy with the lint rules (so long as the 1.9 hash syntax is required).

stve:
🆒  I just gave it a try and was able to auto-correct all violations when using `rubocop-github`, putting the project in good shape. 

I did get a few warnings related to rubocop renaming some things:

```
Style/BlockEndNewline has the wrong namespace - should be Layout
Style/EndOfLine has the wrong namespace - should be Layout
Style/InitialIndentation has the wrong namespace - should be Layout
Style/SpaceAfterMethodName has the wrong namespace - should be Layout
Style/SpaceAfterColon has the wrong namespace - should be Layout
Style/SpaceAfterComma has the wrong namespace - should be Layout
Style/SpaceAfterNot has the wrong namespace - should be Layout
Style/SpaceAfterSemicolon has the wrong namespace - should be Layout
Style/SpaceAroundBlockParameters has the wrong namespace - should be Layout
Style/SpaceAroundEqualsInParameterDefault has the wrong namespace - should be Layout
Style/SpaceInsideArrayPercentLiteral has the wrong namespace - should be Layout
Style/SpaceInsideBrackets has the wrong namespace - should be Layout
Style/SpaceInsideParens has the wrong namespace - should be Layout
Style/SpaceInsideRangeLiteral has the wrong namespace - should be Layout
Style/Tab has the wrong namespace - should be Layout
Style/TrailingBlankLines has the wrong namespace - should be Layout
Style/TrailingWhitespace has the wrong namespace - should be Layout
```

My guess is `rubocop-github` was based off an earlier version of `rubocop`, apart from the annoyance of the output, it doesn't result in any violations. I didn't push these changes yet. Shall we roll with it?

oreoshake:
Well that's annoying. If you lock rubocop to `~> 0.47` do the warnings go away?

oreoshake:
Based on the CI failures, maybe now would be a good time to prep for 4.x and drop EOL'd versions from Travis.

stve:
@oreoshake adding a pessimistic version constraint on `~> 0.47.0` removed the warnings. Using `~> 0.47` was too optimistic as it allowed minor versions `> 0.47`.

There's another hiccup though, CI fails on Ruby 2.0.0 and Ruby 1.9.3 due to the following:

```
rubocop-github was resolved to 0.1.0, which depends on
  ruby (>= 2.3.0)
```

As [1.9.3][193eol] and [2.0.0][200eol] have reached end of life, we can probably drop support for both at this point, though probably with a major version bump?

[193eol]: https://www.ruby-lang.org/en/news/2015/02/23/support-for-ruby-1-9-3-has-ended/
[200eol]: https://www.ruby-lang.org/en/news/2016/02/24/support-plan-of-ruby-2-0-0-and-2-1/

stve:
> maybe now would be a good time to prep for 4.x and drop EOL'd versions from Travis.

👍  Agreed, looks like [2.1.0 has also just reached EOL][210eol] as well

[210eol]: https://www.ruby-lang.org/en/news/2017/04/01/support-of-ruby-2-1-has-ended/

oreoshake:
@stve merge master and this should become green

stve:
@oreoshake all green now 💚

Rakefile
```diff
@@ -1,28 +1,37 @@
...
-require 'net/http'
-require 'net/https'
+# frozen_string_literal: true
+require "bundler/gem_tasks"
+require "rspec/core/rake_task"
+require "net/http"
+require "net/https"
 
 desc "Run RSpec"
+# RSpec::Core::RakeTask.new(:spec)
```
oreoshake: What's this for?

stve: Oops! 🍌  I was trying to make running `bundle exec rake` run both rspec and rubocop and left this by accident

btw, I think we could probably move `--format progress` to `.rspec` and remove the block here.

oreoshake: > btw, I think we could probably move --format progress to .rspec and remove the block here.

:+1:

stve: Fixed now. 

FYI, I removed the `desc "Run Rspec"` line as well as it overrode the description it inherits from RSpec.

Issue #286, opened 2016-09-06
secure_headers 4.0


oreoshake:
I want to start working on a list of breaking changes for the new version. The idea is to make it harder to do the wrong thing.
- [x] Require `script-src` to be set. Missing `script-src`s inherit from `default-src`. In general `default-src` lists are far too permissive for something like `script-src`.
- [x] Require `object-src` to be set. See ⬆️ 
- [x] Default cookies to secure, httponly, samsite=lax
- [x] Yeah, a search for "deprecated" turns up a few more things we can remove entirely.
- [x] Notice of intent to drop support for <= 2.1 (2.1 EOL'd on April 1, 2017)

stve:
Not really a change per se, but we should remove `SecureHeaders::Configuration#secure_cookies=` as it was deprecated in 3.2.0 and replaced with `SecureHeaders::Configuration#cookies=`

oreoshake:
Yeah, a [search for "deprecated"](https://github.com/twitter/secureheaders/search?utf8=%E2%9C%93&q=deprecated) turns up a few more things we can remove entirely.

oreoshake:
Notice of intent to drop support for <= 2.1 ([2.1 EOL'd on April 1, 2017](https://www.ruby-lang.org/en/news/2017/04/01/support-of-ruby-2-1-has-ended/))

oreoshake:
The PR to remove < 2.1 support is up [here](https://github.com/twitter/secureheaders/pull/333).

oreoshake:
> - Require `script-src` to be set. Missing `script-src`s inherit from `default-src`. In general `default-src` lists are far too permissive for something like `script-src`.
> - Require `object-src` to be set. See ⬆️ 

These changes are trivial validation changes that should be easy for anyone looking for an easy PR.

> - Default cookies to secure, httponly, samsite=lax

This should be as simple as changing [these three lines](https://github.com/twitter/secureheaders/blob/master/lib/secure_headers/headers/cookie.rb#L19-L22) ⁉️

oreoshake:
Another change I want in 4.0: a better default policy. Currently, the default policy is `default-src https:` which is pretty good but I'd want to at least set `object-src` to none, add `data:` to fonts and images, `unsafe-inline` to `style-src`, and a few others. E.g. https://github.com/hanami/hanami/blob/23b9562f44e1185ab561bab842af0cb8ff217671/spec/support/shared_examples/cli/generate/app.rb#L272

[...]

PR #340, opened 2017-07-12
[4.x prep] Require a script-src, change default CSP

oreoshake:
This handles the first two items from https://github.com/twitter/secureheaders/issues/286

> - Require `script-src` to be set. Missing `script-src`s inherit from `default-src`. In general `default-src` lists are far too permissive for something like `script-src`.
> - Require `object-src` to be set. See ⬆️ 

/cc @stve @jacobbednarz

CI isn't green yet, but it works on my machine

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,8 @@
+## 4.x
+
+- `script_src` is required to be set in CSP configs. Falling back to *any* `default-src` can be bad. It's certainly possible for this to not cause a problem but better safe than sorry.
+- The default CSP has been changed: `default-src 'self', form-action 'self'; object-src 'none'; script-src https:; style-src 'self' 'unsafe-inline' https:` This policy is "more secure" and more reasonable.
+
 ## 3.6.5
 
 Update clear-site-data header to use current format specified by the specification.
--- a/Gemfile
+++ b/Gemfile
@@ -17,6 +17,6 @@ end
 
 group :guard do
   gem "growl"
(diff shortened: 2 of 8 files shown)


oreoshake:
Bah, hits submit and immediately realizes `Require object-src to be set. See ⬆️` was not implemented 😢

stve:
This might fly in the face of a previous approach but would it makes sense to make the default policy `none` by default and let the implementor loosen those policies as they see fit? On one hand, it creates more friction "out of the box", but on the other, if you are implementing CSP it forces one to really consider each and every policy. Food for thought...

If the intention is to provide a reasonable best practice as a default then this looks good to me 👍

oreoshake:
> default policy none by default and let the implementor loosen those policies as they see fit?

I'm torn but I've consistently advocated for reasonable defaults in frameworks. This was immensely successful when twitter moved to SOA/scala. A few frameworks (not rails) have been taking this route too.

Anecdotally, knowledge about CSP is very limited. Something that just works :tm: is more universally understandable. But maybe this feeling is out of date? I have a hard time thinking objectively about this.

jacobbednarz:
I'm pretty happy with the new defaults. I don't think it's overly permissive yet will provide a good starting base out of the box. Everyone I have spoken to about CSP _wants_ to get serious but doesn't know decent defaults and I think (at least for Rails) this is a great start!

oreoshake:
I made `img-src` more lenient to account for `data:` images and favicons (in addition to `https:`).

PR #341, opened 2017-07-14
Adds secure, httponly, and SameSite=lax to cookies by default

anglinb:
Adds `secure` and `httponly` flags to cookies by default, targeting a 4.0 release tracked by #286. 

/CC @stve @oreoshake 
### Outstanding Questions:

Should 4.0 also include `SameSite=Lax`? 

> [- Default cookies to secure, httponly, samsite=lax](https://github.com/twitter/secureheaders/issues/286#issuecomment-305678390)

### Suggested CHANGELOG entry
(I'm not sure if I should add this to the actual CHANGELOG file b/c I don't know which release this will get pulled into)

> Adds `secure` and `httponly` flags to all cookies by default.

## All Prs

- [x] Has tests
- [x]  Documentation updated

### DIFF
--- a/lib/secure_headers/headers/cookie.rb
+++ b/lib/secure_headers/headers/cookie.rb
@@ -2,6 +2,7 @@
 require "cgi"
 require "secure_headers/utils/cookies_config"
 
+
 module SecureHeaders
   class CookiesConfigError < StandardError; end
   class Cookie
@@ -14,12 +15,16 @@ module SecureHeaders
 
     attr_reader :raw_cookie, :config
 
+    COOKIE_DEFAULTS = {
+      httponly: true,
+      secure: true,
+    }.freeze
+
     def initialize(cookie, config)
       @raw_cookie = cookie
       unless config == SecureHeaders::OPT_OUT
         config ||= {}
-        config[:secure] = true if config[:secure].nil?
-        config[:httponly] = true if config[:httponly].nil?
+        config = COOKIE_DEFAULTS.merge(config)
       end
       @config = config
       @attributes = {
--- a/lib/secure_headers/headers/cookie.rb
+++ b/lib/secure_headers/headers/cookie.rb
@@ -98,7 +98,7 @@ module SecureHeaders
     end
 
     def flag_samesite?
-      return false if config == OPT_OUT or config[:samesite] == OPT_OUT
+      return false if config == OPT_OUT || config[:samesite] == OPT_OUT
       flag_samesite_lax? || flag_samesite_strict?
     end
 
--- a/lib/secure_headers/headers/cookie.rb
+++ b/lib/secure_headers/headers/cookie.rb
@@ -16,6 +16,11 @@ module SecureHeaders
 
     def initialize(cookie, config)
       @raw_cookie = cookie
+      unless config == SecureHeaders::OPT_OUT
+        config ||= {}
+        config[:secure] = true if config[:secure].nil?
+        config[:httponly] = true if config[:httponly].nil?
+      end
       @config = config
       @attributes = {
         httponly: nil,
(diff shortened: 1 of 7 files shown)


oreoshake:
@stve you're the most familiar with this code. Can you take a look :eyes:?

spec/lib/secure_headers/headers/cookie_spec.rb
```diff
@@ -5,37 +5,42 @@ module SecureHeaders
   describe Cookie do
     let(:raw_cookie) { "_session=thisisatest" }
 
-    it "does not tamper with cookies when unconfigured" do
-      cookie = Cookie.new(raw_cookie, {})
+    it "does not tamper with cookies when using OPT_OUT is used" do
```
stve: 👍 for adding support for `SecureHeaders::OPT_OUT` this was a HUGE miss on my part in the initial implementation.

lib/secure_headers/headers/cookie.rb
```diff
@@ -16,6 +16,11 @@ def validate_config!(config)
 
     def initialize(cookie, config)
       @raw_cookie = cookie
+      unless config == SecureHeaders::OPT_OUT
+        config ||= {}
```
stve: I wonder if it would make sense to define the default as `{ secure: true, httponly: true}` as opposed to checking for `nil` in the configuration.

I'm of the opinion that if a configuration is defined (even if it disables `secure` or `httponly`), you shouldn't have to opt-out of each configuration explicitly.

oreoshake:
> Main question is how we want to handle a configuration that doesn't specify `secure` or `httponly`.

You should have to explicitely say `secure: OPT_OUT` and `http_only: OPT_OUT` in my mind.

stve:
> You should have to explicitly say secure: OPT_OUT and http_only: OPT_OUT in my mind.

@anglinb can you add tests for when `secure` and `httponly` are set with `OPT_OUT`? After that I think we're good to merge.

lib/secure_headers/headers/cookie.rb
oreoshake: Per https://github.com/twitter/secureheaders/pull/341#issuecomment-315867482, I think we should go with

```ruby
config[:secure] = true unless config[:secure] == OPT_OUT
config[:httponly] = true unless config[:httponly] == OPT_OUT
```

oreoshake:
@anglinb sorry for not identifying the `OPT_OUT` stuff during :pear:. Do you have time to update this? I can take over if you're busy but I'm hoping to get 4.x out this week or next.

anglinb:
@oreoshake @stve Thanks for taking a look! I'll address your feedback this afternoon--totally missed these notifications 🙃

oreoshake:
While you're at it... let's default `samesite` to `lax` 

/me ducks

Break all the things. I'll add a post-install message to broadcast this change (along with the others).

anglinb:
@oreoshake Sounds good. Will take care of the two changes:

- [ ] Force users to explicitly opt out by supplying `{http_only:  OPT_OUT}`
- [ ] Add `SameSite=Lax` by default

I'm still going to allow the global, OPT_OUT. So if `config = OPT_OUT` is equivalent to `config = {http_only: OPT_OUT, secure: OPT_OUT, same_site: OPT_OUT}`. Does that make sense? @stve @oreoshake

oreoshake:
> I'm still going to allow the global, OPT_OUT. So if config = OPT_OUT is equivalent to config = {http_only: OPT_OUT, secure: OPT_OUT, same_site: OPT_OUT}. Does that make sense?

:+1:

anglinb:
^ Just added `SameSite=Lax`. Please take a look when you have a sec. I _think_ addressed all of the comments. 😄

@stve @oreoshake

PR #346, opened 2017-07-25
Prevent global cookie OPT_OUT from blowing up in middleware

oreoshake:
When we started requiring the use of `OPT_OUT` for cookie settings to mean "don't touch my cookies," the tests didn't catch the case where the secure flag cannot be set on non-https requests.

When `config[:secure]` executes, it calls the `[]` function which is wired to go `boom`. This bypasses that code block if the value of `config` is `OPT_OUT`.

/cc @anglinb

### DIFF
--- a/lib/secure_headers.rb
+++ b/lib/secure_headers.rb
@@ -24,7 +24,7 @@ module SecureHeaders
   class NoOpHeaderConfig
     include Singleton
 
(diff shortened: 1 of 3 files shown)


anglinb:
Whoops! Good catch. Thanks @oreoshake!

PR #354, opened 2017-09-05
Distinguish between source lists directives and non-source list directives

ptoomey3:
The prior behavior tends to default treat directives that are lists as "source lists". But, a source list kind of has specific behavior (ex. they support `*` to mean "all"). So, there are not meaningful for other lists such as `media-type` and `sandbox`. These directives have "lists", but they are unique to their usage. So, we shouldn't treat them like a source list. This PR adds support for `media-type` and `sandbox`. In particular it treats them as follows:

* `sandbox` - Can set a maximally strict policy using either `sandbox: true` or `sandbox: []`. Any other policy requires each element in the list to be of the form `allow-*`.
* `media-type` - Requires that each element in the list be of the form `type/subtype`. 

In addition the the above, I also (I think) simplified some of the surrounding logic. I bumped into places where the existing logic had to be adjusted to account for the new top level directive types anyway. And, as I was refactoring, I think I was able to remove some of the nestedness of what was there.

### DIFF
--- a/lib/secure_headers/headers/policy_management.rb
+++ b/lib/secure_headers/headers/policy_management.rb
@@ -116,18 +116,6 @@ module SecureHeaders
     # everything else is in between.
     BODY_DIRECTIVES = ALL_DIRECTIVES - [DEFAULT_SRC, REPORT_URI]
 
-    # These are directives that do not inherit the default-src value. This is
-    # useful when calling #combine_policies.
-    NON_FETCH_SOURCES = [
-      BASE_URI,
-      FORM_ACTION,
-      FRAME_ANCESTORS,
-      PLUGIN_TYPES,
-      REPORT_URI
-    ]
-
-    FETCH_SOURCES = ALL_DIRECTIVES - NON_FETCH_SOURCES
-
     VARIATIONS = {
       "Chrome" => CHROME_DIRECTIVES,
       "Opera" => CHROME_DIRECTIVES,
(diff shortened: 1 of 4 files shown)


lib/secure_headers/headers/policy_management.rb
```diff
@@ -264,28 +268,35 @@ def ua_to_variation(user_agent)
...
+          sandbox_list?(directive) ||
+          media_type_list?(directive)
+      end
+
       # For each directive in additions that does not exist in the original config,
       # copy the default-src value to the original config. This modifies the original hash.
       def populate_fetch_source_with_default!(original, additions)
         # in case we would be appending to an empty directive, fill it with the default-src value
         additions.each_key do |directive|
-          if !original[directive] && ((source_list?(directive) && FETCH_SOURCES.include?(directive)) || nonce_added?(original, additions))
```
ptoomey3: This is the bit I think I simplified. But, a second set of eyes to make sure I didn't remove some underlying assumption would be ✨

lib/secure_headers/headers/policy_management.rb
```diff
@@ -302,38 +313,77 @@ def nonce_added?(original, additions)
...
       # source expression.
-      def validate_directive!(directive, source_expression)
+      def validate_directive!(directive, value)
+        ensure_valid_directive!(directive)
         case ContentSecurityPolicy::DIRECTIVE_VALUE_TYPES[directive]
         when :boolean
-          unless boolean?(source_expression)
-            raise ContentSecurityPolicyConfigError.new("#{directive} must be a boolean value")
-          end
-        when :string
```
ptoomey3: I don't think I saw any `:string` types so I removed it here.

lib/secure_headers/headers/policy_management.rb
```diff
@@ -302,38 +313,77 @@ def nonce_added?(original, additions)
...
+        end
+      end
+
+      # Private: validates that a sandbox token expression:
+      # 1. is an array of strings or optionally `true` (to enable maximal sandboxing)
+      # 2. For arrays, each element is of the form allow-*
+      def validate_sandbox_expression!(directive, sandbox_token_expression)
+        # We support sandbox: true to indicate a maximally secure sandbox.
+        return if boolean?(sandbox_token_expression) && sandbox_token_expression == true
+        ensure_array_of_strings!(directive, sandbox_token_expression)
```
ptoomey3: If someone sent something bogus like `sandbox: false` we would technically catch that here. But, we never actually reach this code if you do that, since there is some higher level code that does something like `next unless value` for all directives. I didn't want to open pandora's box by removing that higher level assumption.

lib/secure_headers/headers/policy_management.rb
```diff
@@ -264,28 +268,35 @@ def ua_to_variation(user_agent)
...
-          if !original[directive] && ((source_list?(directive) && FETCH_SOURCES.include?(directive)) || nonce_added?(original, additions))
-            if nonce_added?(original, additions)
-              inferred_directive = directive.to_s.gsub(/_nonce/, "_src").to_sym
-              unless original[inferred_directive] || NON_FETCH_SOURCES.include?(inferred_directive)
-                original[inferred_directive] = default_for(directive, original)
-              end
-            else
-              original[directive] = default_for(directive, original)
-            end
+          directive = if nonce_added?(original, additions)
```
oreoshake: We can get rid of `nonce_added?` and just check that `directive.to_s.end_with?("_nonce")`. It's pretty dirty that nonces are essentially added as extra directives. Probably best to clean that up, but in a separate PR.

lib/secure_headers/headers/policy_management.rb
oreoshake: 🤔 now that we use the `OPT_OUT` pattern that should probably be revisited

ptoomey3: You caught me..I noticed this as well and punted :-). I'll take a quick pass at addressing it.

ptoomey3:
@oreoshake - Took a pass at the changes you asked for in https://github.com/twitter/secureheaders/pull/354/commits/36a22c82dcf6cf5dc0ab4a719785516d8fedfbba.

PR #356, opened 2017-09-16
Updates to documentation to clarify the default configuration misunderstanding.

srisa:
This PR contains
1) Word of caution about configuration.(A lot of people including me misunderstood the given configuration to be default configuration)
2) Added relevant documentation to use clear_site_data headers on specific requests.

### DIFF
--- a/README.md
+++ b/README.md
@@ -55,7 +55,7 @@ If you do not supply a `default` configuration, exceptions will be raised. If yo
 
 All `nil` values will fallback to their default values. `SecureHeaders::OPT_OUT` will disable the header entirely.
 
-**Word of caution:**  The following is not a default configuration per se. It serves as a reference to all the available options. You should read more about these headers and determine what is appropriate for your requirements. 
+**Word of caution:**  The following is not a default configuration per se. It serves as a sample implementation of the configuration. You should read more about these headers and determine what is appropriate for your requirements. 
 
 ```ruby
 SecureHeaders::Configuration.default do |config|
--- a/README.md
+++ b/README.md
@@ -55,6 +55,8 @@ If you do not supply a `default` configuration, exceptions will be raised. If yo
 
 All `nil` values will fallback to their default values. `SecureHeaders::OPT_OUT` will disable the header entirely.
 
+**Word of caution:**  The following is not a default configuration per se. It serves as a reference to all the available options. You should read more about these headers and determine what is appropriate for your requirements. 
+
 ```ruby
 SecureHeaders::Configuration.default do |config|
   config.cookies = {
--- a/docs/per_action_configuration.md
+++ b/docs/per_action_configuration.md
@@ -113,7 +113,7 @@ class ApplicationController < ActionController::Base
   # Configuration override to send the Clear-Site-Data header.
   SecureHeaders::Configuration.override(:clear_browser_cache) do |config|
     config.clear_site_data = [
-      SecureHeaders::ClearSiteData::ALL
+      SecureHeaders::ClearSiteData::ALL_TYPES
     ]
   end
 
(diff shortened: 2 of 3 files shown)


oreoshake:
@srisa words cannot explain how thankful I am. Given that my oversight has led to this problem, I would love it if others could chime in on this change.

docs/per_action_configuration.md
```diff
@@ -103,3 +103,32 @@ Content-Security-Policy: ...
...
+## Clearing browser cache
+
+You can clear the browser cache after the logout request by using the following.
+
+``` ruby
+class ApplicationController < ActionController::Base
+  # Configuration override to send the Clear-Site-Data header.
+  SecureHeaders::Configuration.override(:clear_browser_cache) do |config|
+    config.clear_site_data = [
+      SecureHeaders::ClearSiteData::ALL
```
reedloden: ALL_TYPES, perhaps?

docs/per_action_configuration.md
```diff
@@ -103,3 +103,32 @@ Content-Security-Policy: ...
...
+  # header.
+  #
+  # Returns nothing.
+  def clear_browser_cache
+    SecureHeaders.use_secure_headers_override(request, :clear_browser_cache)
+  end
+end
+
+class SessionsController < ApplicationController
+  after_filter  :clear_browser_cache, only: :destroy
```
reedloden: after_action is more common nowadays

README.md
```diff
@@ -55,6 +55,8 @@ If you do not supply a `default` configuration, exceptions will be raised. If yo
 
 All `nil` values will fallback to their default values. `SecureHeaders::OPT_OUT` will disable the header entirely.
 
+**Word of caution:**  The following is not a default configuration per se. It serves as a reference to all the available options. You should read more about these headers and determine what is appropriate for your requirements.
```
jacobbednarz: > It serves as a reference to all the available options.

This isn't quite true since not all of the directives have all available options. For example, HSTS:

```rb
# Add "; preload" and submit the site to hstspreload.org for best protection.
config.hsts = "max-age=#{20.years.to_i}; includeSubdomains"
```

Perhaps wording along the lines of "It serves as a sample implementation of the configuration"? Alternatively, we could go through and add all the possible values however I think this opens a bigger can of worms for those who don't read this warning and just copy-pasta into their app and all of a sudden, their site is bricked because of incorrectly configured HSTS or HPKP.

jacobbednarz: We could also just treat those dangerous directives differently and allow them to not be the full available options so people cannot shoot themselves in the foot without knowing it.

docs/per_action_configuration.md
```diff
@@ -103,3 +103,32 @@ Content-Security-Policy: ...
...
+end
+
```
jacobbednarz: Looks like you've got an extra newline here that we could probably get rid of to keep the markdown examples pretty.

srisa:
I addressed the comments and sent in an update!

oreoshake:
I absolutely *love* the changes in `docs/per_action_configuration.md` but it feels like the disclaimer in `README.md` is not enough. 

I think a 🚨 :fire: :hocho: with a few `<hr>` tags or (something as drastic) are needed. I might even suggest that the block in question not live in the parent `README` but in some separate subdocument ("global customization").

chrislewis60:
Looks good ... but I agree that the warning alone won't cover all bases and it doesn't specifically mention  that it's the clear_site_data that's the one in the defaulkt config that will break all forms in Chrome v61 for example. For what it's worth, I'd like to see clear_site_data separated off from the main default config and discussed serarately on the homepage, so that what remains is a relatively safe starting point. Thanks.

oreoshake:
:cool: let's iterate from here.

PR #361, opened 2017-09-24
Use `SecureHeaders` namespace for OPT_OUT docs

jacobbednarz:
The documentation for upgrading to 4.0 suggests using `OPT_OUT` for
cookie values you don't wish to set however this throws a `NameError`

```
NameError: uninitialized constant OPT_OUT
```

To correctly resolve the `OPT_OUT` value, we need to use the
`SecureHeaders` namespace.


jacobbednarz:
agggghhh, you're not going to believe that i missed the line below this that has the same error 🙄

PR #364, opened 2017-10-02
Support worker-src CSP directive

### LINKED ISSUE #357: Support worker-src
`worker-src` should be supported in the CSP header.

It is supported by Chrome 56+.

Documentation on MDN: https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Security-Policy/worker-src
SPec: https://w3c.github.io/webappsec-csp/#directive-worker-src

arr-dev:
Closes #357 

I'm not sure about Opera support.
`SecureHeaders::PolicyManagment::VARIATIONS` defines same directives in Chrome and Opera, 
but MDN says that Opera doesn't support it.
 
Other than that, I believe everything else is in order.

### DIFF
--- a/README.md
+++ b/README.md
@@ -95,6 +95,7 @@ SecureHeaders::Configuration.default do |config|
     plugin_types: %w(application/x-shockwave-flash),
     script_src: %w('self'),
     style_src: %w('unsafe-inline'),
+    worker_src: %w('self'),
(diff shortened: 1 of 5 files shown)


oreoshake:
Released in 4.0.1 (maybe that should have been 4.1.0 🤔) and backported to 3.7.2

Thanks!

PR #367, opened 2017-10-19
Update Expect-CT to use separator in spec

ptoomey3:
## All PRs:

* [x] Has tests
* [x] Documentation updated

## Update Expect-CT to use separator specified in the current spec

It seems like semicolons were maybe specified at some point, since various articles written (ex. https://scotthelme.co.uk/a-new-security-header-expect-ct/) use semicolon as the delimiter. But, in the current spec (https://tools.ietf.org/html/draft-ietf-httpbis-expect-ct), commas are used. 

/cc @oreoshake - Since we just deployed this, I think I can confirm the semi-colons aren't working, as I didn't see any expect-ct related data when querying chrome://net-internals/#hsts.

### DIFF
--- a/spec/lib/secure_headers/headers/expect_certificate_transparency_spec.rb
+++ b/spec/lib/secure_headers/headers/expect_certificate_transparency_spec.rb
@@ -3,13 +3,13 @@ require "spec_helper"
 
 module SecureHeaders
   describe ExpectCertificateTransparency do
-    specify { expect(ExpectCertificateTransparency.new(max_age: 1234, enforce: true).value).to eq("enforce; max-age=1234") }
+    specify { expect(ExpectCertificateTransparency.new(max_age: 1234, enforce: true).value).to eq("enforce, max-age=1234") }
     specify { expect(ExpectCertificateTransparency.new(max_age: 1234, enforce: false).value).to eq("max-age=1234") }
     specify { expect(ExpectCertificateTransparency.new(max_age: 1234, enforce: "yolocopter").value).to eq("max-age=1234") }
-    specify { expect(ExpectCertificateTransparency.new(max_age: 1234, report_uri: "https://report-uri.io/expect-ct").value).to eq("max-age=1234; report-uri=\"https://report-uri.io/expect-ct\"") }
+    specify { expect(ExpectCertificateTransparency.new(max_age: 1234, report_uri: "https://report-uri.io/expect-ct").value).to eq("max-age=1234, report-uri=\"https://report-uri.io/expect-ct\"") }
     specify do
       config = { enforce: true, max_age: 1234, report_uri: "https://report-uri.io/expect-ct" }
-      header_value = "enforce; max-age=1234; report-uri=\"https://report-uri.io/expect-ct\""
+      header_value = "enforce, max-age=1234, report-uri=\"https://report-uri.io/expect-ct\""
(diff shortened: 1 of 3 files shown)


spec/lib/secure_headers/headers/expect_certificate_transparency_spec.rb
```diff
@@ -3,13 +3,13 @@
```
ptoomey3: I also updated the name of this test, as "expect certificate" was ever so slightly confusing :smile:.

jacobbednarz: Thanks! I think this was my bad from when I had it named `expect_ct_spec` => `expect_certificate_spec` 😛

jacobbednarz:
Nice find @ptoomey3! We've got this deployed using the commas and while I don't see anything in `chrome://net-internals/#hsts` for one of our domains, I also don't see anything for GitHub (confirmed it is in my requests just in case you were being sneaky with the rollout 😉 ). Like the varying degrees of HPKP I would expect to see _something_ however I'm getting sweet nothing.

Is it just me? Or does the net internals not report on unenforced policies for Expect CT?

ptoomey3:
@jacobbednarz - You on Chrome 61 (the first version to ship with it I *think*)? We haven't updated the deploy on github.com to use the patch here just yet, but will post back when we do.

jacobbednarz:
I'm on 64. Derpp, I just re-read (and read again!) to see GitHub is on the same delimiter as us so it is expected to be the same.

ptoomey3:
Calling @estark37 - What should we be looking for on chrome://net-internals/#hsts :smile:?

ptoomey3:
Yeah..I manually forced our policy to use commas using an intercepting proxy, and don't see any results when using the "Query Expect-CT domain" in Chrome 64.

estark37:
Urgh, chrome://net-internals#hsts doesn't seem to be working for report-only policies, maybe? Sorry. I'll have to debug when I'm back on vacation next week. You can try it on alex-gaynor.net to see what it should look like when it's working.
screen shot 2017-10-20 at 8 48 43 am

estark37:
Filed https://bugs.chromium.org/p/chromium/issues/detail?id=776658 to fix chrome://net-internals when I'm back next week. As a side note, a policy injected with an intercepting proxy won't work because, like HPKP, Expect-CT is ignored for invalid certificates or certificates that chain to locally-installed roots.

ptoomey3:
> As a side note, a policy injected with an intercepting proxy won't work because, like HPKP, Expect-CT is ignored for invalid certificates or certificates that chain to locally-installed roots.

I had one of those moments when you wake up and instantly have a thought from something you were doing yesterday. Today it was "Doh..I bet you anything Emily left a comment saing 'it won't work with an intercepting proxy because, like HPKP, ..." :smile:. I'll double check this again once we have the update here deployed.

ptoomey3:
> You can try it on alex-gaynor.net to see what it should look like when it's working.

Cool..yeah, I see that site working fine. I'm guessing all will be well once we push out the semicolon to comma change. Thanks!

ptoomey3:
> Filed https://bugs.chromium.org/p/chromium/issues/detail?id=776658 to fix chrome://net-internals when I'm back next week.

Looks like you can take one task off your todo list next week:

screen shot 2017-10-20 at 9 00 38 am

PR #371, opened 2017-11-14
Update Referrer-Policy to support multiple token values

ptoomey3:
## All PRs:

* [x] Has tests
* [x] Documentation updated

I've yet to hunt down that this is official supported by [the spec](https://www.w3.org/TR/referrer-policy/). But, it seems to be a thing :smile:. Double checking with @estark37 before merging.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,7 @@
+## 5.0.2
+
+- Updates `Referrer-Policy` header to support multiple policy values
+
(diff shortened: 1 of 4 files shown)


oreoshake:
I'd say this change is enough of a corner case mixed with being a somewhat advanced and lesser used item, I'm not going to backport this change.

oreoshake:
Released in 5.0.2

PR #372, opened 2017-11-14
Add nonced versions of Rails link/include tags

steveh:
Adds support for nonced versions of Rails' CSS, JavaScript and Webpacker link/include tags.

* nonced_stylesheet_link_tag
* nonced_javascript_include_tag
* nonced_javascript_pack_tag

## All PRs:

* [x] Has tests
* [x] Documentation updated

### DIFF
--- a/lib/secure_headers/view_helper.rb
+++ b/lib/secure_headers/view_helper.rb
@@ -7,19 +7,43 @@ module SecureHeaders
     class UnexpectedHashedScriptException < StandardError; end
 
     # Public: create a style tag using the content security policy nonce.
-    # Instructs secure_headers to append a nonce to style/script-src directives.
+    # Instructs secure_headers to append a nonce to style-src directive.
     #
     # Returns an html-safe style tag with the nonce attribute.
     def nonced_style_tag(content_or_options = {}, &block)
       nonced_tag(:style, content_or_options, block)
     end
 
+    # Public: create a stylesheet link tag using the content security policy nonce.
+    # Instructs secure_headers to append a nonce to style-src directive.
+    #
+    # Returns an html-safe link tag with the nonce attribute.
+    def nonced_stylesheet_link_tag(*args, &block)
+      stylesheet_link_tag(*args, nonce: content_security_policy_nonce(:style), &block)
+    end
+
     # Public: create a script tag using the content security policy nonce.
-    # Instructs secure_headers to append a nonce to style/script-src directives.
+    # Instructs secure_headers to append a nonce to script-src directive.
     #
     # Returns an html-safe script tag with the nonce attribute.
     def nonced_javascript_tag(content_or_options = {}, &block)
       nonced_tag(:script, content_or_options, block)
+    end
+
+    # Public: create a script src tag using the content security policy nonce.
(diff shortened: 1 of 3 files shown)


lib/secure_headers/view_helper.rb
```diff
@@ -7,21 +7,45 @@ module ViewHelpers
...
+      stylesheet_link_tag(*args, nonce: content_security_policy_nonce(:style), &block)
+    end
+
     # Public: create a script tag using the content security policy nonce.
-    # Instructs secure_headers to append a nonce to style/script-src directives.
+    # Instructs secure_headers to append a nonce to script-src directive.
     #
     # Returns an html-safe script tag with the nonce attribute.
     def nonced_javascript_tag(content_or_options = {}, &block)
       nonced_tag(:script, content_or_options, block)
```
oreoshake: Should just be a passthrough call to `nonced_javascript_include_tag`?

lib/secure_headers/view_helper.rb
```diff
@@ -7,21 +7,45 @@ module ViewHelpers
     class UnexpectedHashedScriptException < StandardError; end
 
     # Public: create a style tag using the content security policy nonce.
-    # Instructs secure_headers to append a nonce to style/script-src directives.
+    # Instructs secure_headers to append a nonce to style-src directive.
     #
     # Returns an html-safe style tag with the nonce attribute.
     def nonced_style_tag(content_or_options = {}, &block)
       nonced_tag(:style, content_or_options, block)
```
oreoshake: Should this be a passthrough to `nonced_stylesheet_link_tag` instead?

spec/lib/secure_headers/view_helpers_spec.rb
```diff
@@ -64,6 +70,16 @@ def content_tag(type, content = nil, options = nil, &block)
     "<#{type}#{options}>#{content}</#{type}>"
   end
 
+  def javascript_include_tag(source, options = {})
+    content_tag(:script, nil, options.merge(src: source))
+  end
+
+  alias_method :javascript_pack_tag, :javascript_include_tag
```
oreoshake: What is the reason for the alias? Can we accomplish the same thing without the metaprogramming?

lib/secure_headers/view_helper.rb
steveh: No, `nonced_style_tag` generates `<style>` whereas `nonced_stylesheet_link_tag` generates a `<link />`

spec/lib/secure_headers/view_helpers_spec.rb
steveh: We could, but for the purpose of this test the method is identical. Would you prefer that?

lib/secure_headers/view_helper.rb
steveh: The Rails API is a little different:

`javascript_tag(content_or_options_with_block = nil, html_options = {}, &block)`

`javascript_include_tag(*sources)`

It might work as a passthrough though, if that's preferred?

oreoshake:
Released in 5.0.3

Sorry for the confusion, I was completely misreading things :smile:

PR #373, opened 2017-11-21
Add missing `nonced_stylesheet_pack_tag`

paulfri:
Equivalents were added in 1e81a0a.

This also fixes a bug preventing passing any arguments to these tags. Tests were updated with a sample to reflect this. I can split that into a separate PR or separate commits if you prefer.

## All PRs:

* [x] Has tests
* [x] Documentation updated

### DIFF
--- a/docs/per_action_configuration.md
+++ b/docs/per_action_configuration.md
@@ -72,6 +72,8 @@ body {
 <%= nonced_javascript_pack_tag "pack.js" %>
 
 <%= nonced_stylesheet_link_tag "link.css" %>
+
+<%= nonced_stylesheet_pack_tag "pack.css" %>
 ```
 
 becomes:
--- a/lib/secure_headers/view_helper.rb
+++ b/lib/secure_headers/view_helper.rb
@@ -19,7 +19,9 @@ module SecureHeaders
     #
     # Returns an html-safe link tag with the nonce attribute.
     def nonced_stylesheet_link_tag(*args, &block)
-      stylesheet_link_tag(*args, nonce: content_security_policy_nonce(:style), &block)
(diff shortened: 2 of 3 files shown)


oreoshake:
I think it's fine to lump in the arg fix but I think this behavior necessitates a test to ensure we don't accidentally break that down the road..

paulfri:
@oreoshake The existing tests fail if you keep the options I added in the example template (`media: :all` and `defer: true`, but use the old implementation.

steveh:
👍  I was about to submit essentially the same PR, but I don't think this (or my PR) will work if you pass a Hash of options, as opposed to kwargs.

Rails uses its own helper method to detect if the last argument is a Hash and extracts options from there. I don't know whether we should replicate it, or just live with only keyword arguments.

https://github.com/rails/rails/blob/7bfc79ec9bb054b97b2f961a75f8c789999b68dc/actionview/lib/action_view/helpers/asset_tag_helper.rb#L75

paulfri:
@steveh Good catch. `**kwargs` should capture the trailing arguments even if they're passed as a hash, so I think it's just a question if we want to merge those options with the sources if multiple are given (versus just support one source at a time). I don't know that I've ever seen `javascript_include_tag` used with multiple sources like in their documentation though.

oreoshake:
> I don't know whether we should replicate it, or just live with only keyword arguments.

Replicating seems less surprising. I wouldn't worry too much about drift in these APIs, but that's a concern.

paulfri:
I pushed up another implementation. I don't know that it's really Better Code™ but it supports passing multiple sources to the helpers, which the previous one did not. Tests were updated to reflect this.

oreoshake:
Released in 5.0.4

PR #382, opened 2018-01-20
unconditionally send nonces and unsafe-inline when working with nonces

### LINKED ISSUE #381: Android 4.4 unsupported nonces
# Bugs

Using a script with added `content_security_policy_script_nonce` tag is not being executed on stock Android 4.4 device and emulator (in stock browser and webview). The same script is loaded on Android 4.2/4.3.

I've tried manually passing `'unsafe-inline'` instead and the script was loaded.

User-Agents:
```
Mozilla/5.0 (Linux; Android 4.4.2; Android SDK built for x86 Build/KK) AppleWebKit/537.36 (KHTML, like Gecko) Version/4.0 Chrome/30.0.0.0 Mobile Safari/537.36
Mozilla/5.0 (Linux; Android 4.4.4; XT1021 Build/KXC21.5-40) AppleWebKit/537.36 (KHTML, like Gecko) Version/4.0 Chrome/33.0.0.0 Mobile Safari/537.36
```

### Expected outcome

The script should be executed.

### Actual outcome

Script with nonce is blocked. The following logs are produced:
```
01-19 12:45:47.009 2241-2241/com.android.browser E/browser: Console: The source list for Content Security Policy directive 'script-src' contains an invalid source: ''nonce-VqttPDPhtEDgC+kNL57SmLkTpCzD+6bPDh2TbhrROJg=''. It will be ignored. :0
01-19 12:45:47.009 2241-2241/com.android.browser E/browser: Console: Refused to execute inline script because it violates the following Content Security Policy directive: "script-src 'self' 'nonce-VqttPDPhtEDgC+kNL57SmLkTpCzD+6bPDh2TbhrROJg='".
```

### Config

Using version 5.0.4:
```ruby
SecureHeaders::Configuration.configure do |config|
  config.csp = {
    :default_src => SecureHeaders::OPT_OUT,
    :script_src => %w('self')
  }
  config.hsts = SecureHeaders::OPT_OUT
  config.x_download_options = SecureHeaders::OPT_OUT
  config.x_frame_options = SecureHeaders::OPT_OUT
  config.x_permitted_cross_domain_policies = SecureHeaders::OPT_OUT
end
```
### Generated headers

```
Content-Security-Policy: script-src 'self' 'nonce-VqttPDPhtEDgC+kNL57SmLkTpCzD+6bPDh2TbhrROJg='
```

oreoshake:
Fixes #381 

/cc @jasssonpet can you see if this resolves your issue?

## All PRs:

* [x] Has tests
* [x] Documentation updated N/A

This pull request stops guessing whether or not a UA supports nonces and relies on the backwards-compatible behavior of sending `'unsafe-inline'` with every `'nonce-`. 

The reason for sniffing was mostly because Safari < 9 would generate console messages about an unknown directive value (`nonce-*`). Safari 9 is two versions behind so in theory fewer people are using it. Given the bugs UA sniffing like this has caused, I'm less sympathetic to those on unsupported browsers craving a clean console.

### DIFF
--- a/lib/secure_headers/headers/content_security_policy.rb
+++ b/lib/secure_headers/headers/content_security_policy.rb
@@ -213,7 +213,7 @@ module SecureHeaders
     # unsafe-inline, this is more concise.
     def append_nonce(source_list, nonce)
       if nonce
-        source_list.push(*["'nonce-#{nonce}'", UNSAFE_INLINE])
+        source_list.push("'nonce-#{nonce}'", UNSAFE_INLINE)
       end
 
       source_list
--- a/lib/secure_headers/headers/content_security_policy.rb
+++ b/lib/secure_headers/headers/content_security_policy.rb
@@ -212,11 +212,7 @@ module SecureHeaders
     # unsafe-inline, this is more concise.
     def append_nonce(source_list, nonce)
       if nonce
-        if nonces_supported?
-          source_list << "'nonce-#{nonce}'"
-        else
-          source_list << UNSAFE_INLINE
-        end
+        source_list.push(*["'nonce-#{nonce}'", UNSAFE_INLINE])
       end
 
       source_list
@@ -254,10 +250,6 @@ module SecureHeaders
       else
         VARIATIONS[OTHER]
       end
-    end
-
-    def nonces_supported?
-      @nonces_supported ||= self.class.nonces_supported?(@parsed_ua)
     end
 
     def symbol_to_hyphen_case(sym)
--- a/docs/upgrading-to-6-0.md
+++ b/docs/upgrading-to-6-0.md
@@ -42,3 +42,9 @@ Prior to 6.0.0 SecureHeaders pre-built and cached the headers that corresponded 
 ## Configuration the default configuration more than once will result in an Exception
 
 Prior to 6.0.0 you could conceivably, though unlikely, have `Configure#default` called more than once. Because configurations are dynamic, configuring more than once could result in unexpected behavior. So, as of 6.0.0 we raise `AlreadyConfiguredError` if the default configuration is setup more than once.
+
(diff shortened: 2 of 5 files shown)


jasssonpet:
Yes, I've tested the proposed changes and can confirm this does solve the issue.

lib/secure_headers/headers/content_security_policy.rb
```diff
@@ -212,11 +212,7 @@ def populate_nonces(directive, source_list)
     # unsafe-inline, this is more concise.
     def append_nonce(source_list, nonce)
```
jasssonpet: You may want to remove the "depending on browser" part from the method comments.

oreoshake:
> Update: Now I see the following warnings in the latest Firefox 57 that may be confusing:

UGH STOP DOING THIS BROWSERS

oreoshake:
> Update: Now I see the following warnings in the latest Firefox 57 that may be confusing:

I didn't see that warning. I upgraded to 58 and I don't see it there either (testing it by visiting twitter.com which uses nonces). I do however see other warnings about how X-Frame-Options is ignored because of frame-ancestors and I'm beginning to think the whack a mole game of keeping a clean console should be abandoned as a goal for this gem. 

I'm pretty sure someone else put a lot of thought into those warnings and they had good reason to do so.

oreoshake:
I think I'll ship this with https://github.com/twitter/secureheaders/pull/383 in v6.0 unless there are objections.

My feeling is that the console will be noisy and it's not worth adding code/bugs for the sake of a clean console. Thinks are built to be backwards compatible when possible so I don't understand why the console has to be so noisy about legitimate use.

oreoshake:
Released in 6.0.0.alpha01

PR #383, opened 2018-01-22
Rewrite the entire notion of named overrides (breaking change)

### LINKED ISSUE #380: Using `override` after using `append_content_security_policy_directives` leads to unexpected behavior
# Bugs

:bug:

### Expected outcome

1. Add CSP changes to the default policy using `append/override_content_security_policy_directives`
1. Use `SecureHeaders.use_secure_headers_override`
1. CSP additions from step 1 are preserved

### Actual outcome

1. CSP additions from step 1 are not preserved

### Example code

```ruby
class ThingController < ApplicationController
  before_action :secure_all_the_things

  SecureHeaders::Configuration.override(:disable_referrer_policy) do |config|
    config.referrer_policy = SecureHeaders::OPT_OUT
  end

  def secure_all_the_things # CSP changes are added and removed here
    append_content_security_policy_directives(script_src: 'foo.com')
    use_secure_headers_override(:disable_referrer_policy)
  end
end
```

@ptoomey3 is working around this issue, but this is surprising behavior. I expected the CSP exceptions to persist even though I'm "overriding" here. Unfortunately, the internals here are a bit messy and the tests are brittle.

Maybe augmenting `named_appends` is a better approach. Also, we could add a `override_referrer_policy` method, similar to `override_x_frame_options`.

ptoomey3:
:warning: Work in progress ⚠️ 

This PR is very likely not yet ready for a full review. But, I wanted to push up the work in progress to get a sense of whether the direction of it is palatable, since I am introducing some relatively foundational changes that will definitely break some existing uses. 

#380 outlines a bug related to the current implementation of named overrides. There is an underlying assumption that named override is more or less a "fork" of a parent configuration. The parent configuration, at the point the override is defined, is duplicated and then the changes in the override are applied. The resulting configuration is statically stored in a class instance variable and can be retrieved when it is requested with `Configuration.get(:named_override`). Unfortunately, this model is directly at odds with how configurations work at runtime. Imagine the following scenario:

* Request is using the default config
* At runtime we append on another `script-src`
* At runtime we apply an override

Currently, the above scenario results in the appended `script-src` from not showing up. The reason is that, when we call the override, it applies the config that was generated when the override was defined. That config is simply a copy of the original default config with the overridden directives applied. So, the dynamic changes made to the default config are lost. 

In order to fix this we, unfortunately, need to redefine the entire notion of an override. Most callers would assume that an override will override the "current configuration" at runtime. As a result, we can't statically computer the override at the point it is defined. Instead, we need to treat overrides more like how we treat named appends. We store the provided block for an override in a hash that maps the name to the `Proc`. Then, when the override is requested, we `instance_eval` the `Proc` with the current configuration for the request. 

Also, during this PR it became evident that trying to cache the generated configs was incredibly complex and prone to error. This PR removes the caching and just computes the headers on demand for each request. We already hit this case a fair bit if one ever used per-request overrides anyway, since the cached values would get invalidated depending on which code path you went through on a given request. The computation overhead of making a header is relatively low and we don't expect it to be a concern. If there is a concern we could optimize the base case and only precompute the default config. But, if any changes are applied, then the precomputation is lost. As I'm typing this it sounds like that could be a nice win, since so few folks probably use dynamic policies. Thoughts?

fixes #380

## All PRs:

* [x] Has tests
* [ ] Documentation updated

### DIFF
--- a/lib/secure_headers.rb
+++ b/lib/secure_headers.rb
@@ -148,6 +148,7 @@ module SecureHeaders
     def header_hash_for(request)
       prevent_dup = true
       config = config_for(request, prevent_dup)
+      config.validate_config!
       user_agent = UserAgent.parse(request.user_agent)
       headers = config.generate_headers(user_agent)
 
@@ -198,7 +199,7 @@ module SecureHeaders
     # Falls back to the global config
     def config_for(request, prevent_dup = false)
       config = request.env[SECURE_HEADERS_CONFIG] ||
-        Configuration.get(Configuration::DEFAULT_CONFIG)
+        Configuration.get
 
 
       # Global configs are frozen, per-request configs are not. When we're not
--- a/lib/secure_headers.rb
+++ b/lib/secure_headers.rb
@@ -166,7 +166,7 @@ module SecureHeaders
     # name - the name of the previously configured override.
     def use_secure_headers_override(request, name)
       config = config_for(request)
-      config.apply_override(name)
+      config.override(name)
       override_secure_headers_request_config(request, config)
     end
 
@@ -195,7 +195,7 @@ module SecureHeaders
     # Falls back to the global config
     def config_for(request, prevent_dup = false)
       config = request.env[SECURE_HEADERS_CONFIG] ||
-        Configuration.get
+        Configuration.default
 
 
       # Global configs are frozen, per-request configs are not. When we're not
--- a/lib/secure_headers.rb
+++ b/lib/secure_headers.rb
@@ -165,13 +165,9 @@ module SecureHeaders
     #
     # name - the name of the previously configured override.
     def use_secure_headers_override(request, name)
-      if override = Configuration.overrides(name)
-        config = config_for(request)
-        config.instance_eval(&override)
-        override_secure_headers_request_config(request, config)
-      else
-        raise ArgumentError.new("no override by the name of #{name} has been configured")
-      end
+      config = config_for(request)
+      config.apply_override(name)
+      override_secure_headers_request_config(request, config)
     end
 
     # Public: gets or creates a nonce for CSP.
--- a/lib/secure_headers.rb
+++ b/lib/secure_headers.rb
@@ -1,5 +1,4 @@
 # frozen_string_literal: true
-require "secure_headers/configuration"
 require "secure_headers/hash_helper"
 require "secure_headers/headers/cookie"
 require "secure_headers/headers/public_key_pins"
@@ -18,6 +17,7 @@ require "secure_headers/railtie"
 require "secure_headers/view_helper"
 require "useragent"
 require "singleton"
+require "secure_headers/configuration"
 
 # All headers (except for hpkp) have a default value. Provide SecureHeaders::OPT_OUT
 # or ":optout_of_protection" as a config value to disable a given header
@@ -52,29 +52,9 @@ module SecureHeaders
   HTTPS = "https".freeze
   CSP = ContentSecurityPolicy
 
-  ALL_HEADER_CLASSES = [
-    ExpectCertificateTransparency,
-    ClearSiteData,
-    ContentSecurityPolicyConfig,
-    ContentSecurityPolicyReportOnlyConfig,
-    StrictTransportSecurity,
-    PublicKeyPins,
-    ReferrerPolicy,
-    XContentTypeOptions,
-    XDownloadOptions,
-    XFrameOptions,
-    XPermittedCrossDomainPolicies,
-    XXssProtection
-  ].freeze
-
-  ALL_HEADERS_BESIDES_CSP = (
-    ALL_HEADER_CLASSES -
-      [ContentSecurityPolicyConfig, ContentSecurityPolicyReportOnlyConfig]
-  ).freeze
-
   # Headers set on http requests (excludes STS and HPKP)
-  HTTP_HEADER_CLASSES =
-    (ALL_HEADER_CLASSES - [StrictTransportSecurity, PublicKeyPins]).freeze
+  HTTPS_HEADER_CLASSES =
+    [StrictTransportSecurity, PublicKeyPins].freeze
 
   class << self
     # Public: override a given set of directives for the current request. If a
@@ -153,7 +133,7 @@ module SecureHeaders
     # Public: opts out of setting all headers by telling secure_headers to use
     # the NOOP configuration.
     def opt_out_of_all_protection(request)
-      use_secure_headers_override(request, Configuration::NOOP_CONFIGURATION)
+      use_secure_headers_override(request, Configuration::NOOP_OVERRIDE)
     end
 
     # Public: Builds the hash of headers that should be applied base on the
@@ -168,27 +148,15 @@ module SecureHeaders
     def header_hash_for(request)
       prevent_dup = true
       config = config_for(request, prevent_dup)
-      headers = config.cached_headers
       user_agent = UserAgent.parse(request.user_agent)
+      headers = config.generate_headers(user_agent)
 
-      if !config.csp.opt_out? && config.csp.modified?
-        headers = update_cached_csp(config.csp, headers, user_agent)
-      end
-
-      if !config.csp_report_only.opt_out? && config.csp_report_only.modified?
-        headers = update_cached_csp(config.csp_report_only, headers, user_agent)
-      end
-
-      header_classes_for(request).each_with_object({}) do |klass, hash|
-        if header = headers[klass::CONFIG_KEY]
-          header_name, value = if klass == ContentSecurityPolicyConfig || klass == ContentSecurityPolicyReportOnlyConfig
-            csp_header_for_ua(header, user_agent)
-          else
-            header
-          end
-          hash[header_name] = value
+      if request.scheme != HTTPS
+        HTTPS_HEADER_CLASSES.each do |klass|
+          headers.delete(klass::HEADER_NAME)
         end
       end
+      headers
     end
 
     # Public: specify which named override will be used for this request.
@@ -196,7 +164,9 @@ module SecureHeaders
     #
     # name - the name of the previously configured override.
     def use_secure_headers_override(request, name)
-      if config = Configuration.get(name)
+      if override = Configuration.overrides(name)
+        config = config_for(request)
+        config.instance_eval(&override)
         override_secure_headers_request_config(request, config)
       else
         raise ArgumentError.new("no override by the name of #{name} has been configured")
@@ -284,35 +254,6 @@ module SecureHeaders
     # Returns the config.
     def override_secure_headers_request_config(request, config)
       request.env[SECURE_HEADERS_CONFIG] = config
-    end
-
-    # Private: determines which headers are applicable to a given request.
-    #
-    # Returns a list of classes whose corresponding header values are valid for
-    # this request.
-    def header_classes_for(request)
-      if request.scheme == HTTPS
-        ALL_HEADER_CLASSES
-      else
-        HTTP_HEADER_CLASSES
-      end
-    end
-
-    def update_cached_csp(config, headers, user_agent)
-      headers = Configuration.send(:deep_copy, headers)
-      headers[config.class::CONFIG_KEY] = {}
-      variation = ContentSecurityPolicy.ua_to_variation(user_agent)
-      headers[config.class::CONFIG_KEY][variation] = ContentSecurityPolicy.make_header(config, user_agent)
-      headers
-    end
-
-    # Private: chooses the applicable CSP header for the provided user agent.
-    #
-    # headers - a hash of header_config_key => [header_name, header_value]
-    #
-    # Returns a CSP [header, value] array
-    def csp_header_for_ua(headers, user_agent)
-      headers[ContentSecurityPolicy.ua_to_variation(user_agent)]
     end
   end
 
--- a/lib/secure_headers.rb
+++ b/lib/secure_headers.rb
@@ -195,7 +195,7 @@ module SecureHeaders
     # Falls back to the global config
     def config_for(request, prevent_dup = false)
       config = request.env[SECURE_HEADERS_CONFIG] ||
-        Configuration.default
+        Configuration.send(:default_config)
 
 
       # Global configs are frozen, per-request configs are not. When we're not
--- a/lib/secure_headers/configuration.rb
+++ b/lib/secure_headers/configuration.rb
@@ -5,7 +5,7 @@ module SecureHeaders
   class Configuration
     DEFAULT_CONFIG = :default
     NOOP_OVERRIDE = "secure_headers_noop_override"
-    class NotYetConfiguredError < StandardError; end
+    class AlreadyConfiguredError < StandardError; end
     class IllegalPolicyModificationError < StandardError; end
     class << self
       # Public: Set the global default configuration.
@@ -14,6 +14,18 @@ module SecureHeaders
       #
       # Returns the newly created config.
       def default(&block)
+        if defined?(@default_config) && !block_given?
+          return @default_config
+        else
+          configure(&block)
+        end
+      end
+
+      def configure(&block)
+        if defined?(@default_config)
+          raise AlreadyConfiguredError, "Policy already configured"
+        end
+
         # Define a built-in override that clears all configuration options and
         # results in no security headers being set.
         override(NOOP_OVERRIDE) do |config|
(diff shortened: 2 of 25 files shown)


lib/secure_headers/headers/content_security_policy_config.rb
```diff
@@ -55,12 +53,10 @@ def directive_value(directive)
       end
     end
 
-    def modified?
-      @modified
-    end
-
     def merge(new_hash)
-      ContentSecurityPolicy.combine_policies(self.to_h, new_hash)
+      new_config = self.dup
```
ptoomey3: @oreoshake - I know we had chatted about the expectations of `merge`. But, given that `merge!` was overwriting existing values, it seems like `merge` should to the same (or `merge!` should do the opposite). It seems like `append` should be used for any kind of "combine these policies" use cases.

spec/lib/secure_headers/headers/policy_management_spec.rb
```diff
@@ -18,7 +18,7 @@ module SecureHeaders
         # (pulled from README)
         config = {
           # "meta" values. these will shape the header, but the values are not included in the header.
-          report_only:  true,     # default: false
+          report_only:  false,
```
ptoomey3: I'm now validating `report_only` inside the actual config object. So, we can't go setting `report_only: true` unless we are going to be validating it with a `ContentSecurityPolicyReportOnlyConfig`.

spec/lib/secure_headers_spec.rb
```diff
@@ -60,6 +60,24 @@ module SecureHeaders
         expect(hash["X-Frame-Options"]).to be_nil
       end
 
+      it "Overrides the current default config if default config changes during request" do
```
ptoomey3: This test exercises the bug in https://github.com/twitter/secureheaders/issues/380.

spec/lib/secure_headers_spec.rb
```diff
@@ -116,7 +131,7 @@ module SecureHeaders
         firefox_request = Rack::Request.new(request.env.merge("HTTP_USER_AGENT" => USER_AGENTS[:firefox]))
 
         # append an unsupported directive
-        SecureHeaders.override_content_security_policy_directives(firefox_request, {plugin_types: %w(flash)})
+        SecureHeaders.override_content_security_policy_directives(firefox_request, {plugin_types: %w(application/pdf)})
```
ptoomey3: Since we are now validating the config before we generate headers we need to use a legit value for `plugin_types`.

oreoshake:
> If there is a concern we could optimize the base case and only precompute the default config. But, if any changes are applied, then the precomputation is lost. As I'm typing this it sounds like that could be a nice win, since so few folks probably use dynamic policies. Thoughts?

My personal opinion:

optimizeelsewhere

ptoomey3:
> My personal opinion:

I'm definitely not looking for an opportunity to add a caching bug if I don't need to. 👍

oreoshake:
I would love some input from people more concerned about performance.

ptoomey3:
@oreoshake some super hacky quick benchmarking on my local machine says that a cached set of headers takes 30ms for 1000 invocations (.03ms/call) vs. 170ms for 1000 invocations uncached (.17ms/call). So, it definitely is more expensive. But, I have no clue if .17ms is something to sweat over (so says 10K libraries that each add on their own .17ms :trollface:).

ptoomey3:
@oreoshake some super hacky quick benchmarking on my local machine says that a cached set of headers takes 30ms for 1000 invocations (.03ms/call) vs. 170ms for 1000 invocations uncached (.17ms/call). So, it definitely is more expensive. But, I have no clue if .17ms is something to sweat over (so says 10K libraries that each add on their own .17ms :trollface:).

oreoshake:
> (so says 10K libraries that each add on their own .17ms :trollface: * TO EVERY REQUEST*

Emphasis added.

Anecdotally, non-caching code that generates headers is pretty common and does well on high traffic websites. I think it's more important for lightweight APIs which have higher RPS than heavy web pages.

jacobbednarz:
Once this is in a usable state (perhaps it already is?) I'm happy to run it within our CI suite to confirm correctness on our end and put it through some benchmarks. When we're happy with that, I can roll this out to our fleet and keep you posted on any performance changes.

On a side note, I ❤️ reviewing PRs that _actually_ use the description field to convey the intention and context. It's amazing. Thanks @ptoomey3 🙇 !

ptoomey3:
@jacobbednarz - I think this code could use a bit more review..but it does currently pass all the CI tests. So, it should work more or less for testing/benchmarking/etc. 

> On a side note, I ❤️ reviewing PRs that actually use the description field to convey the intention and context. It's amazing.

Thanks! I tend to veer toward the side of over-explaining my changes 😄

oreoshake:
@jacobbednarz did you notice the branch name? Patrick wants to jump to secure_headers, and I quote:

> “Secure Headers 10 Snow Leopard”

lib/secure_headers.rb
```diff
@@ -228,7 +195,7 @@ def content_security_policy_style_nonce(request)
     # Falls back to the global config
     def config_for(request, prevent_dup = false)
       config = request.env[SECURE_HEADERS_CONFIG] ||
-        Configuration.get(Configuration::DEFAULT_CONFIG)
+        Configuration.get
```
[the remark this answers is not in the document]
oreoshake: :+1: to renaming it

lib/secure_headers/headers/content_security_policy_config.rb
oreoshake: I think we should move as far away from treating this like a hash as possible. Using verbs like append/combine/overwrite would be clearer. But we'll have to alias and deprecate first.

lib/secure_headers/headers/policy_management.rb
```diff
@@ -201,6 +201,7 @@ module ClassMethods
       # Returns a default policy if no configuration is provided, or a
       # header name and value based on the config.
       def make_header(config, user_agent)
```
oreoshake: why is user_agent required here but optional elsewhere?

jacobbednarz:
I've managed to get some performance metrics out of our application for this change. For these I've used [derailed_benchmarks](https://github.com/schneems/derailed_benchmarks) against the existing gem and the local branch that @ptoomey3 has pushed up.

### Current

```
allocated memory by gem
-----------------------------------
      9231  secure_headers-5.0.4

allocated objects by gem
-----------------------------------
       137  secure_headers-5.0.4

retained memory by gem
-----------------------------------
       187  secure_headers-5.0.4
```

### Proposed

```
allocated memory by gem
-----------------------------------
     18809  secureheaders/lib

allocated objects by gem
-----------------------------------
       282  secureheaders/lib

retained memory by gem
-----------------------------------
       436  secureheaders/lib
```

There is a bit of a difference here (especially double the allocated memory) so I think it's worth confirming whether this acceptable for others. This isn't a deal breaker for me but might be slightly concerning for the Hubbers.

As for real requests, there is little difference to the end user response time (using `benchmark-ips`)

```
[current]   13.396  (± 7.5%) i/s - 67.000  in   5.036455s
[proposed]  13.050  (± 7.7%) i/s - 65.000  in   5.034300s
```

lib/secure_headers/headers/policy_management.rb
ptoomey3: no good reason..only because we always pass it when it is used. But, for consistency it would be fine...eventually this is passed to a method that has a `user_agent = OTHER` default value. I'll make it match the others.

ptoomey3: Maybe I should have it (and the others) match the `user_agent = OTHER` pattern though.

lib/secure_headers/configuration.rb
```diff
@@ -193,35 +205,26 @@ def update_x_frame_options(value)
...
 
     def csp=(new_csp)
-      if new_csp.respond_to?(:opt_out?)
-        @csp = new_csp.dup
+      case new_csp
+      when OPT_OUT
+        @csp = new_csp
+      when ContentSecurityPolicyConfig
+        @csp = new_csp
+      when Hash
```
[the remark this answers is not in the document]
ptoomey3: Meh..I think choosing one or the other feels awkward in different places. For example, during initial config it feels really natural to do:

``` ruby
Configuration.configure do |config|
  config.csp = {
    default_src: %w(https: 'self')
  }
end
```

But, it also feels super natural to do

``` ruby
Configuration.override(:footer) do |config|
  config.csp = config.csp.merge(...)
```

So, I guess I tend to vote to leave it as is.

lib/secure_headers/configuration.rb
```diff
@@ -14,9 +15,21 @@ class << self
       #
       # Returns the newly created config.
       def default(&block)
-        config = new(&block)
-        add_noop_configuration
-        add_configuration(DEFAULT_CONFIG, config)
+        if defined?(@default_config)
```
ptoomey3: In the spirit of preserving the integrity of the default config I added this. It is my assumption that a caller should only ever being calling this configuration block a single time, correct? If it is called more than once it implies that we could be running into strange runtime behavior, since the call order could go something like:

``` ruby
Configuration.default
...
some runtime override
...
Configuration.default
```

lib/secure_headers/configuration.rb
```diff
@@ -104,6 +77,17 @@ def deep_copy(config)
         end
       end
 
+      # Private: Returns the internal default configuration. This should only
+      # ever be called by internal callers (or tests) that know the semantics
+      # of ensuring that the default config is never mutated and is dup(ed)
+      # before it is used in a request.
+      def default_config
```
ptoomey3: This private class method makes tests uglier, but it seems worthwhile to preserve hiding the actual default config from public callers.

lib/secure_headers/headers/content_security_policy.rb
```diff
@@ -13,6 +13,7 @@ class ContentSecurityPolicy
     FALLBACK_VERSION = ::UserAgent::Version.new("0")
 
     def initialize(config = nil, user_agent = OTHER)
+      user_agent ||= OTHER
```
ptoomey3: Since we are defaulting `user_agent` to `nil` in various places I thought it made sense to also override `user_agent` if `nil` ends up being passed.

spec/lib/secure_headers/headers/policy_management_spec.rb
```diff
@@ -3,6 +3,11 @@
 
 module SecureHeaders
   describe PolicyManagement do
+    before(:each) do
+      reset_config
```
ptoomey3: We call this in more tests now that it is an exception to reconfigure the default configuration.

spec/lib/secure_headers/headers/policy_management_spec.rb
```diff
@@ -142,17 +147,33 @@ module SecureHeaders
...
+      it "allows report_only to be set in a report-only config" do
+        expect do
+          ContentSecurityPolicy.validate_config!(ContentSecurityPolicyReportOnlyConfig.new(default_opts.merge(report_only: true)))
+        end.to_not raise_error
+      end
     end
 
     describe "#combine_policies" do
+      before(:each) do
+        reset_config
```
ptoomey3: For groups of tests that want to play with the default configuration we need to reset it so we don't get an `AlreadyConfiguredError`

spec/lib/secure_headers/headers/policy_management_spec.rb
```diff
@@ -183,7 +205,8 @@ module SecureHeaders
         non_default_source_additions = ContentSecurityPolicy::NON_FETCH_SOURCES.each_with_object({}) do |directive, hash|
           hash[directive] = %w("http://example.org)
         end
-        combined_config = ContentSecurityPolicy.combine_policies(Configuration.get.csp.to_h, non_default_source_additions)
+        default_policy = Configuration.send(:default_config)
```
ptoomey3: Yup..this makes a handful of tests uglier, but I don't like making private internal data more visible simply for the sake of tests.

spec/spec_helper.rb
```diff
@@ -34,18 +35,21 @@ def expect_default_values(hash)
...
+  expect(hash[SecureHeaders::PublicKeyPins::HEADER_NAME]).to be_nil
 end
 
 module SecureHeaders
   class Configuration
     class << self
-      def clear_configurations
-        @configurations = nil
+      def clear_default_config
+        remove_instance_variable(:@default_config) if defined?(@default_config)
```
ptoomey3: It isn't enough to `nil` out the default config anymore since we are checking `defined?(@default_config)` when we throw an `AlreadyConfiguredError`.

lib/secure_headers/configuration.rb
oreoshake: > It is my assumption that a caller should only ever being calling this configuration block a single time, correct?

There were guards in place to prevent doing exactly that. Not sure if they survived the change.

ptoomey3: But don't the tests do a bunch of `Configuration.default {...}` between tests? That implies it was ok to reconfigure the default config a bunch of times, no?

ptoomey3:
@oreoshake - If you would take a quick read through the 6.0.0 upgrade docs in https://github.com/twitter/secureheaders/pull/383/commits/c0750dadb13005e94d9e295d98fa6e3012fca504 just let me know if there is something I missed, some idiom I didn't follow, or some more detailed guidance you think we should provide.

ptoomey3:
All the unit tests for GitHub pass with these changes. But, I did notice I had refactor a few one-off call sites (used for local scripts used to generate CSP values for static files) in a less than glamorous way:

Before: 

``` ruby
SecureHeaders::CSP.make_header(
  SecureHeaders::Configuration.get(:static_file_policy).csp
)
```

After:

``` ruby
SecureHeaders::CSP.make_header(
  SecureHeaders::Configuration.send(:default_config).dup.override(:static_file_policy).csp
)
```

That is a pretty gross, but is admittedly a fringe usage of this whole library. So, we can either just accept that very few folks will want to do this OR we could add back in a method for fetching the default config. But, unlike the prior `Configuration.get` method, we could always return a `dup` of the policy. Thoughts?

oreoshake:
So hiding `.send(:default_config).dup.override(:static_file_policy)` in `#get`? Sounds good to me.

ptoomey3:
> So hiding .send(:default_config).dup.override(:static_file_policy) in #get? Sounds good to me.

I would just be hiding the `.send(:default_config).dup` bit such that thew new code would like like `Configuration.get.override(:static_file_policy)`. That said..since the prior `Configuration.get` didn't `dup`, what is the lesser evil. Should we kill off `Configuration.get` and replace it with `Configuration.default_dup` or leave it and just change the behavior out from underneath people. I'm somewhat nervous to do that, as it might superficially "work", but come bite people in the but when the realize it is a dup.

oreoshake:
Hmm I guess without any actual information on how people are using this it seems that obvious breakage would be better than surprising breakage. :+1: to `Configuration.default_dup` just to remove a `send` in our code.

ptoomey3:
Ok..I added a `Configuration.dup` in https://github.com/twitter/secureheaders/pull/383/commits/b82b56d6a6d4a78857be7217d06c0aa191f6df07. I chose `dup` as opposed to `default_dup` (or whatever) since, especially as of this PR, the `Configuration` class really does represent a single solitary configuration in addition to a bunch of "named overrides" and "named additions". There is only one configuration..so there probably isn't much need to to use the term "default" anymore. For example, `Configuration.default` is setting the single configuration to its default values. So, `Configuration.dup` felt natural for meaning "give me a dup of the config". Let me know if you think it too ambiguous.

ptoomey3:
I deployed this PR on a lab deployment of github.com and it looked good. The only change I had to make across the entire codebase was:

before:
``` ruby
SecureHeaders::Configuration.get(:static_file_policy).csp
```

after
``` ruby
SecureHeaders::Configuration.dup.override(:static_file_policy).csp
```

As was noted somewhere above, we use the above in a script to manually generate a static file policy that we copy/paste into our static 404, 500, etc files. It is a pretty obscure usage of the library and I don't expect just about anyone is directly operating on configurations like this. Besides the above, no other changes were needed to get our passing and a deployment operational. I'm guessing that means most consumer of the library won't need to change any code🤞.

oreoshake:
Nice! Let's let this bake for a while in production. In the meantime, I'll cut a deprecation release for the upcoming API changes.

jacobbednarz:
This passes CI on our side with flying colours! I'll canary tomorrow and let you know if I hit any weirdness.

ptoomey3:
I added a test and did a branch deploy of the PR I originally had deployed that unearthed https://github.com/twitter/secureheaders/issues/380. Everything seems to work correctly. I'll probably do another deploy to take a closer look/test at all endpoints that declare overrides later.

oreoshake:
I'm going to merge this in to get moving on a new release with a proper alpha since it seems to work even if not yet fully in production.

oreoshake:
Released in 6.0.0.alpha01

PR #386, opened 2018-02-09
Deprecate Configuration#get

oreoshake:
In prep for https://github.com/twitter/secureheaders/pull/383, this will deprecate the use of `Configuration#get` which is not meant to be a public API.

Because it's used internally all over, I added a flag to whitelist its use. I use the flag in tests to avoid any rails env checking.

This should only warn if application code is calling `get`. In the next version, we will instead expose a `dup` method that can be used [as described in @ptoomey3's comment](https://github.com/twitter/secureheaders/pull/383#issuecomment-363858684)

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -19,5 +19,5 @@ group :guard do
   gem "growl"
   gem "guard-rspec", platforms: [:ruby_19, :ruby_20, :ruby_21, :ruby_22, :ruby_23, :ruby_24]
   gem "rb-fsevent"
-  gem 'terminal-notifier-guard'
+  gem "terminal-notifier-guard"
 end
--- a/Gemfile
+++ b/Gemfile
@@ -19,4 +19,5 @@ group :guard do
   gem "growl"
   gem "guard-rspec", platforms: [:ruby_19, :ruby_20, :ruby_21, :ruby_22, :ruby_23, :ruby_24]
   gem "rb-fsevent"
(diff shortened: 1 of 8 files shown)


jacobbednarz:
Do we want to add some test coverage to ensure that when a non-internal `Configuration#get` is called a deprecation  notice is received?

Gemfile
```diff
@@ -19,4 +19,5 @@ group :guard do
   gem "growl"
   gem "guard-rspec", platforms: [:ruby_19, :ruby_20, :ruby_21, :ruby_22, :ruby_23, :ruby_24]
   gem "rb-fsevent"
+  gem 'terminal-notifier-guard'
```
jacobbednarz: Was this intended to be committed?

spec/lib/secure_headers/configuration_spec.rb
```diff
@@ -12,6 +12,11 @@ module SecureHeaders
       expect(Configuration.get(Configuration::DEFAULT_CONFIG, internal: true)).to_not be_nil
     end
 
+    it "warns when using deprecated internal-ish #get API" do
+      expect(Kernel).to receive(:warn).once.with(/`#get` is deprecated/)
+      Configuration.get(Configuration::DEFAULT_CONFIG)
```
jacobbednarz: nice!

PR #389, opened 2018-03-25
Avoid calling content_security_policy_nonce internally

eugeneius:
Rails 5.2 adds support for configuring a Content-Security-Policy header, including adding nonces to tags produced by the `javascript_tag` helper.

Unfortunately, Rails and this gem now both define a helper named `content_security_policy_nonce`:

https://github.com/rails/rails/blob/v5.2.0.rc2/actionpack/lib/action_controller/metal/content_security_policy.rb#L44
https://github.com/twitter/secureheaders/blob/v5.0.5/lib/secure_headers/view_helper.rb#L69

The Rails helper wins over the Secure Headers one, and helpers like `nonced_javascript_tag` currently raise this error on Rails 5.2:

    ArgumentError: wrong number of arguments (given 1, expected 0)

By using a method with a different name internally, we avoid clashing with the Rails implementation, and `nonced_javascript_tag` works again.

### DIFF
--- a/lib/secure_headers/view_helper.rb
+++ b/lib/secure_headers/view_helper.rb
@@ -19,7 +19,7 @@ module SecureHeaders
     #
     # Returns an html-safe link tag with the nonce attribute.
     def nonced_stylesheet_link_tag(*args, &block)
-      opts = extract_options(args).merge(nonce: content_security_policy_nonce(:style))
+      opts = extract_options(args).merge(nonce: _content_security_policy_nonce(:style))
 
       stylesheet_link_tag(*args, opts, &block)
     end
@@ -37,7 +37,7 @@ module SecureHeaders
     #
     # Returns an html-safe script tag with the nonce attribute.
     def nonced_javascript_include_tag(*args, &block)
-      opts = extract_options(args).merge(nonce: content_security_policy_nonce(:script))
+      opts = extract_options(args).merge(nonce: _content_security_policy_nonce(:script))
 
       javascript_include_tag(*args, opts, &block)
     end
@@ -47,7 +47,7 @@ module SecureHeaders
     #
     # Returns an html-safe script tag with the nonce attribute.
     def nonced_javascript_pack_tag(*args, &block)
-      opts = extract_options(args).merge(nonce: content_security_policy_nonce(:script))
+      opts = extract_options(args).merge(nonce: _content_security_policy_nonce(:script))
 
       javascript_pack_tag(*args, opts, &block)
     end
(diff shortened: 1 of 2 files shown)


eugeneius:
I have no strong opinion on the particular method name, and will happily change it to anything else. 🙂

oreoshake:
This is going to break things 😢. While people should use the tag helpers, I'm sure someone is using the raw value for something. Luckily, 6.0 hasn't been released so we can sneak this in there. It was meant to be a private API to begin with. The name collision is unfortunate but it's a polluted namespace.

oreoshake:
Released in v6.0.0.alpha02.

eugeneius:
Thanks @oreoshake!

I think you may have missed [this line](https://github.com/twitter/secureheaders/blob/v6.0.0.alpha02/lib/secure_headers/view_helper.rb#L77) though: `content_security_policy_nonce` will still work as before on Rails < 5.2. Unless I got something wrong, the only behavioural change here is that the tag helper methods (e.g. `nonced_javascript_tag`) will work correctly on Rails 5.2; everything else stays the same.

paulfri:
@eugeneius We were using `content_security_policy_nonce` directly in a couple cases (some third-party script tags), so this is definitely a breaking change on Rails 5.1.

eugeneius:
I'm not saying that Rails didn't break Secure Headers; I'm saying that this patch makes Secure Headers _mostly_ compatible with Rails 5.2, without affecting its compatibility with earlier versions.

As such I think it's not a breaking change, and could be backported to 5.0.x so that there's a non-prerelease version available that (mostly) works with Rails 5.2.0 when it's out in a couple of weeks.

paulfri:
It may be backwards-compatible with Rails (in that `nonced_javascript_tag` works on both Rails 5.1 and 5.2), but it's definitely not backwards compatible with some Rails applications, since a public-accessible API `content_security_policy_nonce` is being renamed (and is now inferred to be private with the prefixed `_`).

eugeneius:
`content_security_policy_nonce` is still available [as an alias](https://github.com/twitter/secureheaders/blob/v6.0.0.alpha02/lib/secure_headers/view_helper.rb#L77) to the underscore-prefixed version. Rails 5.1 applications will still be able to call it, just like before. If calling `content_security_policy_nonce` doesn't work in your app when using v6.0.0.alpha02, please tell me about the error so that I can fix it!

paulfri:
I completely missed that, sorry! We already updated to use the tag helpers (and 6.0.0.alpha02) but it looks like it's indeed backwards-compatible and could be backported to 5.x.

Issue #384, opened 2018-01-24
Remove all UA sniffing?


oreoshake:
UA sniffing is a tricky beast. While it can help optimize security, it's very error prone. See https://github.com/twitter/secureheaders/issues/381 for a recent example. CSP, when implemented correctly, is backwards compatible and somewhat safe on modern browsers.

It sounds like safari is set on keeping a static UA (locking all future versions to 11.2): https://twitter.com/rmondello/status/956236581987172354 I don't think Safari is making the right move here and while I could be entirely wrong, we cannot rely on it for feature detection going forward for modern browsers. **This could be an issue for people who have to support older browsers**.

I don't think this will break anything (on modern browsers). Safari 10 has good CSP 2 support. Chrome (and Opera) and Firefox support is similar enough. Edge/IE don't break in any unexpected way.

### :fire: UA sniffing? :fire: 

It's only there to cover up behavior from browsers from 3+ years or so ago in some cases. 

**If this introduces a few console messages, would everyone come at me with pitchforks**?

oreoshake:
For the record, this would be the second time UA sniffing would be ripped out entirely.

jacobbednarz:
I don't care for console warnings but the reporting endpoint spam might be troublesome (especially for folks for report-uri.io).

> For the record, this would be the second time UA sniffing would be ripped out entirely.

Third time's a charm, right?

oreoshake:
We could automatically throttle reporting for safari oh wait that would require UA sniffing

paulfri:
Similar to #381, I discovered today from a bug report that `script-src 'strict-dynamic'` is broken in Vivaldi because secure_headers isn't sending the nonce in the CSP header. I considered filing a separate bug, but if you're getting rid of UA sniffing, then the backwards-compatible approach would be a reasonable fix as well.

oreoshake:
@paulfri have you upgrade to 6.0.0.alpha01? That should be fixed in the latest alpha

paulfri:
Yep, fixed in the alpha, I'll just upgrade ASAP. Thanks.

oreoshake:
I've created a PR for this: https://github.com/twitter/secureheaders/pull/390

PR #390, opened 2018-04-11
Remove all useragent sniffing

oreoshake:
Fixes https://github.com/twitter/secureheaders/issues/384

I was trying to convert my app to use `frame-src` instead of `child-src` but found the UA sniffing getting in the way. No more magic around the way `child-src` and `frame-src` are handled.

This will result in more console errors but I'm not sure if it will actually cause any breakage. I think the long chain of fallback rules will save us here.


oreoshake:
With some light testing this resulted in one console warning in safari (`worker-src`) and one less warning on firefox (discouraging child-src in favor of `frame/worker-src`).

docs/upgrading-to-6-0.md
```diff
@@ -48,3 +48,7 @@ Prior to 6.0.0 you could conceivably, though unlikely, have `Configure#default`
...
 
 The first example: we will now send `'unsafe-inline'` along with nonce source expressions. This will generate warnings in some consoles but is 100% valid use and was a design goal of CSP in the early days. The concept of versioning CSP lost out and so we're left with backward compatibility as our only option.
+
+## No more frame-src/child-src magic
+
+First there was frame-src. Then there was child-src which deprecated frame-src. Then child-src became deprecated in favor of frame-src and worker-src. In the meantime, every browser did something different. For a while, it was recommended to set child-src and frame-src but the values had to be identical. secure_headers would sniff the UA to determine whether to use frame or child src. That can lead to confusing things like setting frame-src but seeing child-src. If the child-src and frame-src did not match up, an error is raised. This can be very confusing when using dynamic overrides ("Do we use child-src or frame-src?" => :boom:). Now that the dust has settled, I think we can stop sniffing UAs and just go with a straightforward application.
```
jacobbednarz: :tada:

lib/secure_headers/headers/content_security_policy.rb
```diff
@@ -219,13 +185,13 @@ def append_nonce(source_list, nonce)
       source_list
     end
 
-    # Private: return the list of directives that are supported by the user agent,
+    # Private: return the list of directives,
```
jacobbednarz: Do you want to rewrap this line?

jacobbednarz:
Looks reaaaaal nice 💯

lib/secure_headers/headers/content_security_policy.rb
oreoshake: Heh, I should probably take a pass over all of the comments. I'm sure there's plenty of outdated info.

lib/secure_headers/headers/policy_management.rb
```diff
@@ -199,9 +153,9 @@ module ClassMethods
       #
       # Returns a default policy if no configuration is provided, or a
       # header name and value based on the config.
-      def make_header(config, user_agent = nil)
+      def make_header(config)
```
oreoshake: Removing the `user_agent` argument is a breaking change that will require a deprecation release.

PR #391, opened 2018-04-20
Add `gem 'secure_headers'` to README

Dorian:
I thought it was `secureheaders` instead of `secure_headers` (I looked in the `gemspec` :) )


oreoshake:
@dorian I've updated the repo name to match the gem name. 

When secure_headers was first released it pretty much didn't follow any conventions. This repo rename is waaaay overdue (🤞 that the repo rename didn't break anything).

Issue #403, opened 2019-06-25
Do you have to append unsafe-inline?


will:
I just tried to switch from 'self' and domains to 'strict-dynamic' and nonces for the tags, however when doing that, this code 

``` ruby
    # Private: adds a nonce or 'unsafe-inline' depending on browser support.
    # If a nonce is populated, inline content is assumed.
    #
    # While CSP is backward compatible in that a policy with a nonce will ignore
    # unsafe-inline, this is more concise.
    def append_nonce(source_list, nonce)
      if nonce
        source_list.push("'nonce-#{nonce}'", UNSAFE_INLINE)
      end

      source_list
    end
```
https://github.com/twitter/secure_headers/blob/647dcc4c4dd281f0a0744e64da917dbe298994d3/lib/secure_headers/headers/content_security_policy.rb#L180

appended 'unsafe-inline'. I get that it works and the browsers will ignore the unsafe-inline, but I'd be more comfortable if it wasn't there at all. Is there any way to avoid this?

oreoshake:
It would require a code change. I'm not opposed to making that change but it's there for good reason and building an opt-out option would be a good idea.

will:
I'm assuming that the good reason is for backwards compatibility? If that is the case, for my particular site, I'm fine forcing all of my customers to be on up-to-date browsers.

oreoshake:
> I'm assuming that the good reason is for backwards compatibility?

Correct, any clients that don't understand the new directive source will fallback to the `'unsafe-inline`' and those that do understand it, ignore the `'unsafe-inline'`.

will:
For anyone finding this issue in the future, it has been implemented in #404 .

oreoshake:
I'm trying to push a new gem but I think rubygems is having issues? I'll try again later.

```html
<?xml version="1.0" encoding="utf-8"?>
<!DOCTYPE html PUBLIC "-//W3C//DTD XHTML 1.0 Strict//EN"
 "http://www.w3.org/TR/xhtml1/DTD/xhtml1-strict.dtd">
<html>
  <head>
    <title>502 Bad Gateway</title>
  </head>
  <body>
    <h1>Error 502 Bad Gateway</h1>
    <p>Bad Gateway</p>
    <h3>Guru Mediation:</h3>
    <p>Details: cache-bur17539-BUR 1561580744 879826524</p>
    <hr>
    <p>Varnish cache server</p>
  </body>
</html>
```

PR #404, opened 2019-06-25
Add option to disable appending 'unsafe-inline' when using nonces

will:
This is in reference to #403

### DIFF
--- a/lib/secure_headers/headers/content_security_policy.rb
+++ b/lib/secure_headers/headers/content_security_policy.rb
@@ -179,7 +179,8 @@ module SecureHeaders
     # unsafe-inline, this is more concise.
     def append_nonce(source_list, nonce)
       if nonce
-        source_list.push("'nonce-#{nonce}'", UNSAFE_INLINE)
+        source_list.push("'nonce-#{nonce}'")
+        source_list.push(UNSAFE_INLINE) unless @config[:disable_nonce_backwards_compatibility]
       end
 
       source_list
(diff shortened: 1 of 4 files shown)


lib/secure_headers/headers/content_security_policy.rb
```diff
@@ -179,7 +179,8 @@ def populate_nonces(directive, source_list)
     # unsafe-inline, this is more concise.
     def append_nonce(source_list, nonce)
       if nonce
-        source_list.push("'nonce-#{nonce}'", UNSAFE_INLINE)
+        source_list.push("'nonce-#{nonce}'")
+        source_list.push(UNSAFE_INLINE) unless @config[:disable_append_unsafe_inline]
```
oreoshake: Let's call this `disable_nonce_backwards_compatibility` since `append` is somewhat overloaded in this library.

oreoshake:
Thanks for the PR @will! Apologies for those annoying linter errors. https://github.com/twitter/secure_headers/pull/405 should address that. I'll get that merge in so we can update this for nice green checks instead of red ❌

lib/secure_headers/headers/content_security_policy.rb
will: Sure, good idea. I forced pushed the previous commit with this renamed.

will:
Thanks, and thanks for doing a release :)

It also made me think it'd be cool if when a pull request is released, it notified the people in the pull request. Hey @neovintage stop whatever you're working on now and make my idea here happen soon, thanks in advance!

ghiculescu:
Should this be added to the readme?

ghiculescu:
@ghiculescu yes, it should.

ghiculescu:
@ghiculescu Okay great. Made a PR here: https://github.com/twitter/secure_headers/pull/409

Issue #418, opened 2020-01-21
Filtering CSP entries to prevent bypassing rules


mvgijssel:
Consider the following Rails controller action which overwrites the frame ancestors based on some user input:

```ruby
def show
  user_input_domain1 = URI.parse "https://google.com;script-src"
  user_input_domain2 = URI.parse "https://*;.;"
  user_input_domains = [user_input_domain1, user_input_domain2]

  override_content_security_policy_directives(frame_ancestors: whitelisted_domains)
end
```

This results into the following response header:
```
frame-ancestors: https://google.com;script-src *;
```

This shows unexpected output, because by setting the frame ancestors the user is able to change the `script-src` opening possibilities for XSS. 

One solution to this would be to filter out the CSP rules inside of specific CSP rules:

```ruby
frame_ancestor = value.gsub('script-src','').gsub('img-src', '')
```

oreoshake:
Thanks for the report @mvgijssel. This is something @gregose had brought up a loooong time ago wrt to policy "injection" via semicolons.

I think the better approach would be to escape or raise errors upon seeing a semicolon mid-directive. I believe it would be better to raise an error but that could be breaking change, requiring a deprecation and major version bump. 

It's easy to say "don't do that" but what you appear to be doing is absolutely supported. Accepting user input into a policy seems to bypass the whole point of CSP, but sometimes a single non-perfect opt-out can help make the other 99% of the application safer until things can be rearchitected.

I should have fixed this a looooooong time ago, tech debt!

oreoshake:
This might be worth a security advisory as well.

bwillis:
> It's easy to say "don't do that" but what you appear to be doing is absolutely supported. Accepting user input into a policy seems to bypass the whole point of CSP, but sometimes a single non-perfect opt-out can help make the other 99% of the application safer until things can be rearchitected.

In the end your call, but it's surprising to me to be able to set a different CSP directive when you are explicitly stating `frame_ancestors`. I would expect that the library would prevent you from abusing it in this way, regardless of where the input is coming from. Do you have any examples of a legitimate reason to use it in this way?

oreoshake:
I :100: agree that this behavior should be fixed. There's no legitimate reason that I can think of.

oreoshake:
I'm just waiting on the backport to finish building. I'll push the fixed versions for 3.x, 5.x, and 6.x and issue an advisory after.

bwillis:
[...]

PR #422, opened 2020-01-23
Move semicolon/newline handling to validation and raise errors

oreoshake:
A followup to https://github.com/twitter/secure_headers/issues/418, instead of safely escaping bad values reject them at time of use. This applies to static configurations at boot time and dynamic configurations as well. All policies flow through the validation.

In a breaking release, we can start raising errors when encountering these misconfigurations and/or malicious attempts.

In adding additional validation, I tightened up the assertions a bit so we don't have tests passing when they shouldn't.

### DIFF
--- a/lib/secure_headers/headers/policy_management.rb
+++ b/lib/secure_headers/headers/policy_management.rb
@@ -164,6 +164,10 @@ module SecureHeaders
 
     REQUIRE_SRI_FOR_VALUES = Set.new(%w(script style))
 
+    # These values are not valid in a policy and can be used to escape context
+    # boundaries
+    ESCAPE_SEQUENCE = /(\n|;)/
+
     module ClassMethods
       # Public: generate a header name, value array that is user-agent-aware.
       #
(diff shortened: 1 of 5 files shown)


lib/secure_headers/headers/policy_management.rb
```diff
@@ -325,10 +329,10 @@ def validate_sandbox_expression!(directive, sandbox_token_expression)
         return if boolean?(sandbox_token_expression) && sandbox_token_expression == true
         ensure_array_of_strings!(directive, sandbox_token_expression)
         valid = sandbox_token_expression.compact.all? do |v|
-          v.is_a?(String) && v.start_with?("allow-")
+          v.is_a?(String) && v.start_with?("allow-") && v !~ ESCAPE_SEQUENCE
```
oreoshake: It would probably be cleaner to move this check to `ensure_array_of_strings`.

But the belt-and-suspenders approach of also raising an error on output is also probably a good idea.

jobertabma:
@oreoshake RE #418: we were notified that adding characters like `'$()` to the CSP makes certain browser ignore the header. Is fixing this something you can consider when you finish this PR?

oreoshake:
Ugh, ya. That probably deserves another advisory.

oreoshake:
If someone else is willing to push this to completion, I’d be happy to review and release. I don’t see myself getting to this anytime soon.

jobertabma:
@oreoshake Happy to get this over the finish line. I'll pick up another issue first to go through the codebase and then I'll come back to this one (until someone beats me to it).

oreoshake:
The validation is all over the place, and adding more will work but I’ve been wondering if a real programmer could use the actual grammar from the spec or something close enough.

jobertabma:
@oreoshake I'll start working on this now. Will read the spec (send positive thoughts!) and see how we can improve the validation. I'll look through your code first, but I might publish a separate branch to explore alternative approaches.

PR #436, opened 2020-03-26
Raise on override defined config

### LINKED ISSUE #435: Overriding the same override leads to ???
# Bugs

Using `override(:foo)` with the same `:foo` multiple times probably shouldn't be supported and feels a bit like wandering into undefined behavior.

### Expected outcome

I guess I expected it to raise an error.

### Actual outcome

🤷‍♂ didn't dig in

### Config

```ruby
    SecureHeaders::Configuration.override(:preview_policy) do |config|
      #...
    end

    SecureHeaders::Configuration.override(:preview_policy) do |config|
      # a different ...
    end
```

### Generated headers

🤷‍♂ didn't dig in. This feels non-controversial enough.

jobertabma:
## All PRs:

* [x] Has tests
* [x] Documentation updated

This PR changes the behavior of the `override` and `named_append` methods to raise an exception when it is called more than once with the same configuration name. This is to avoid unexpected behavior in case a configuration is used in some way in between two configuration declarations.

Fixes #435

### DIFF
--- a/docs/named_overrides_and_appends.md
+++ b/docs/named_overrides_and_appends.md
@@ -1,6 +1,6 @@
 ## Named Appends
 
-Named Appends are blocks of code that can be reused and composed during requests. e.g. If a certain partial is rendered conditionally, and the csp needs to be adjusted for that partial, you can create a named append for that situation. The value returned by the block will be passed into `append_content_security_policy_directives`. The current request object is passed as an argument to the block for even more flexibility.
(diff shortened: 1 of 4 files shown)


oreoshake:
> **Open question**: should the `named_append` method also check the existence of a configuration in the `@overrides` hash and vice versa?

🤔 it'd probably be confusing and I can't think of a reason why you'd want overrides and appends to have the same name, or at least I think you'd be OK with making the required change once your app stopped booting?

jobertabma:
@oreoshake I've updated the PR based on your comment, should be ready for review. Thanks!

PR #463, opened 2021-02-05
use un-rolled method for duplciating configs instead of serializing to and from hash

oreoshake:
To my surprise, this seems to have dubious performance benefits. The `dup`s were happening via `write_attribute`...


oreoshake:
The unrolling of this method, as opposed to reflection, feels dubious. But the lack of serialization to and from hashes seems like a good idea?

PR #465, opened 2021-02-05
Add missing CSP version 3 directives

ggalmazor:
## All PRs:

* [x] Has tests
* [x] Documentation updated

## Adding a new CSP directive

  - Chrome, Edge, and Opera support them
  - Safari, Firefox, and IE **don't** support these directives yet
* What does it do?
  - More info about these at https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Security-Policy/script-src-attr
  - These headers should allow any value that's valid for their parent directives `script_src` and `style_src`
  - https://w3c.github.io/webappsec-csp/#directive-script-src-attr
  - https://w3c.github.io/webappsec-csp/#directive-script-src-elem
  - https://w3c.github.io/webappsec-csp/#directive-style-src-attr
  - https://w3c.github.io/webappsec-csp/#directive-style-src-elem

### DIFF
--- a/README.md
+++ b/README.md
@@ -75,7 +75,11 @@ SecureHeaders::Configuration.default do |config|
     sandbox: true, # true and [] will set a maximally restrictive setting
     plugin_types: %w(application/x-shockwave-flash),
     script_src: %w('self'),
+    script_src_elem: %w('self'),
+    script_src_attr: %w('self'),
     style_src: %w('unsafe-inline'),
+    style_src_elem: %w('unsafe-inline'),
+    style_src_attr: %w('unsafe-inline'),
     worker_src: %w('self'),
     upgrade_insecure_requests: true, # see https://www.w3.org/TR/upgrade-insecure-requests/
     report_uri: %w(https://report-uri.io/example-csp)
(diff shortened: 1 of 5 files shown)


oreoshake:
✨ hello @ggalmazor, this is a great addition! Would you be interested in adding a test showing that the newly added directives are included in a generated policy? I think we'd need that for extra assurance.

ggalmazor:
Hi, @oreoshake! Yes, I'll be glad to add that spec to the PR.

ggalmazor:
Is there any other similar spec that I could use for inspiration?

ggalmazor:
For now, I've added some specs to `content_security_policy_spec.rb` but let me know if you'd want me to move them elsewhere, or if you'd expect another way of testing this :)

oreoshake:
That's great, thank you @ggalmazor. I'll get this merged and see if I can potentially batch it up for a release with any of my pending pull requests or just release this on its own in the next few days.

ggalmazor:
Thanks, @oreoshake. That sounds perfect either way :)

oreoshake:
Released in 6.3.2

PR #475, opened 2022-01-17
Add global config for nonce (and hash) application

### LINKED ISSUE #470: Support nonced tags while using report only mode
# Feature Requests

### Problem

We're running a fairly large Rails app with a lot of hard coded script tags, and like most people we're using the **report only** feature of CSP to ensure we don't break functionality while moving forward. In order to reduce the number of reports generated we would like to fix these inline script ocurrences with nonced tags (at least temporarily).
However, as soon as we use `nonced_javascript_tag` on a page, the `nonce` directive appears on the **enforcing** CSP header and this breaks all other inline tags because the browser ignores the `unsafe_inline` directive.

### Solution Ideas

* `nonced_javascript_tag, report_only: true`
* `report_only_nonced_javascript_tag`
* Break backwards compatibility and stop including the `nonce` directive automatically upon usage of the helper tags.

pcasaretto:
* [X] Has tests
* [ ] Documentation updated

Fixes #470

Hi there!

Sorry for the long hiatus.
I'm opening this as is to get feedback on the proposal.
I've added a test case for my current scenario and it is currently passing.

However, I'm unsure what to do for some other scenarios.
I've put togueter a table with all possible combinations and marked the ones I'm unsure with "?"
This is assuming a call was made to `content_security_policy_script_nonce`

| CSP Config | Report Only Config | Apply Config | Result                     |
|------------|--------------------|--------------|----------------------------|
| PRESENT    | PRESENT            | nil          | nonce added to both        |
| PRESENT    | PRESENT            | :both        | nonce added to both        |
| PRESENT    | PRESENT            | :enforced    | nonce added to enforce     |
| PRESENT    | PRESENT            | :report_only | nonce added to report_only |
| PRESENT    | OPT_OUT            | nil          | nonce added to enforce     |
| PRESENT    | OPT_OUT            | :both        | ?                          |
| PRESENT    | OPT_OUT            | :enforced    | nonce added to enforce     |
| PRESENT    | OPT_OUT            | :report_only | ?                          |
| OPT_OUT    | PRESENT            | nil          | nonce added to report_only |
| OPT_OUT    | PRESENT            | :both        | ?                          |
| OPT_OUT    | PRESENT            | :enforced    | ?                          |
| OPT_OUT    | PRESENT            | :report_only | nonce added to report_only |
| OPT_OUT    | OPT_OUT            | nil          | ?                          |
| OPT_OUT    | OPT_OUT            | :both        | ?                          |
| OPT_OUT    | OPT_OUT            | :enforced    | ?                          |
| OPT_OUT    | OPT_OUT            | :report_only | ?                          |

### DIFF
--- a/lib/secure_headers.rb
+++ b/lib/secure_headers.rb
@@ -2,6 +2,7 @@
 require "secure_headers/hash_helper"
 require "secure_headers/headers/cookie"
 require "secure_headers/headers/content_security_policy"
+require "secure_headers/headers/content_security_policy_nonce_apply_to"
 require "secure_headers/headers/x_frame_options"
 require "secure_headers/headers/strict_transport_security"
 require "secure_headers/headers/x_xss_protection"
@@ -208,7 +209,7 @@ module SecureHeaders
 
     def config_and_target(request, target)
       config = config_for(request)
-      target = guess_target(config) unless target
+      target ||= config.csp_nonces_applied_to || guess_target(config)
       raise_on_unknown_target(target)
       [config, target]
     end
@@ -262,8 +263,8 @@ module SecureHeaders
     SecureHeaders.opt_out_of_header(request, header_key)
   end
 
-  def append_content_security_policy_directives(additions)
-    SecureHeaders.append_content_security_policy_directives(request, additions)
+  def append_content_security_policy_directives(additions, target)
+    SecureHeaders.append_content_security_policy_directives(request, additions, target)
   end
 
   def override_content_security_policy_directives(additions)
--- a/lib/secure_headers/configuration.rb
+++ b/lib/secure_headers/configuration.rb
@@ -126,6 +126,7 @@ module SecureHeaders
       expect_certificate_transparency: ExpectCertificateTransparency,
       csp: ContentSecurityPolicy,
       csp_report_only: ContentSecurityPolicy,
+      csp_nonces_applied_to: ContentSecurityPolicyNonceApplyTo,
       cookies: Cookie,
     }.freeze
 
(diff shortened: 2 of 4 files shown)


pcasaretto:
ping @oreoshake

oreoshake:
Apologies, this fell off my radar. I'd have to dig around a little more to make sure there wouldn't be side effects to these decisions and I remember having a little trouble keeping the edge cases in mind here. But at a quick glance, this would be my first guess at desired behavior.


> CSP Config	Report Only Config	Apply Config	Result

> PRESENT	OPT_OUT	:report_only	?
> OPT_OUT	PRESENT	:enforced	?
> OPT_OUT	OPT_OUT	:both	?
> OPT_OUT	OPT_OUT	:enforced	?
> OPT_OUT	OPT_OUT	:report_only	?

Raise an error? These entries all seem contradictory.

> PRESENT	OPT_OUT	:both	?
> OPT_OUT	PRESENT	:both	?

Raise an error? While these are partial contradictions, it could lead to surprising behavior. I would also accept a noop. Something about `:both` being partially inaccurate seems to be worse than e.g. `OPT_OUT OPT_OUT :enforced` which doesn't seem to make a lot of sense in my head.

> OPT_OUT	OPT_OUT	nil	?

Noop? No lies were told here. Because the default value is `nil`, raising an error in this case would require someone to opt out of CSP three times 😄 

Again, apologies if I'm way off here but have we considered having two separate values that would both need to be set rather than this report_only/enforced/both situation? I think that would give us an a strong case for raising errors for all of these question marks.

Thanks for the issue, PR, and followup @pcasaretto. This is back on my radar.

pcasaretto:
For transparency: I've left the company I was working for and this is no longer a priority for me.
I'm closing it for now.

PR #477, opened 2022-03-30
normalize domains with trailing slashes

keithamus:
## All PRs:

* [x] Has tests
* [ ] Documentation updated - N/A

## Adding a new header

* What does it do?

## Adding a new CSP directive

* What does it do?

---

This PR normalises CSP source expressions to exclude trailing slashes from the Domain _if_ there is no other information in the path.

CSP3 more explicitly calls this out in the [path match algorithm](https://www.w3.org/TR/CSP/#path-part-match):

> If path A consists of one character that is equal to the U+002F
> SOLIDUS character (/) and path B is empty, return "Matches".

Also a URL like `example.com/foo` will match a source expression of `example.com`, as well as `example.com/`, so having two source expressions listed like this is redundant.

### DIFF
--- a/lib/secure_headers/headers/content_security_policy.rb
+++ b/lib/secure_headers/headers/content_security_policy.rb
@@ -125,6 +125,7 @@ module SecureHeaders
       else
         source_list = populate_nonces(directive, source_list)
         source_list = reject_all_values_if_none(source_list)
+        source_list = normalize_uri_paths(source_list)
 
         unless directive == REPORT_URI || @preserve_schemes
           source_list = strip_source_schemes(source_list)
@@ -146,6 +147,19 @@ module SecureHeaders
         source_list
       end
     end
+
+    def normalize_uri_paths(source_list)
+      source_list.map do |source|
+        # Normalize domains ending in a single / as without omitting the slash accomplisheg the same.
+        # https://www.w3.org/TR/CSP3/#match-paths § 6.6.2.10 Step 2
+        if source.chomp("/").include?("/")
+          source
+        else
+          source.chomp("/")
+        end
+      end
+    end
+
 
     # Removes duplicates and sources that already match an existing wild card.
     #
--- a/spec/lib/secure_headers/headers/content_security_policy_spec.rb
+++ b/spec/lib/secure_headers/headers/content_security_policy_spec.rb
@@ -48,9 +48,17 @@ module SecureHeaders
         expect(csp.value).to eq("default-src * 'unsafe-inline' 'unsafe-eval' data: blob:")
       end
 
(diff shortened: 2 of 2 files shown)


lib/secure_headers/headers/content_security_policy.rb
```diff
@@ -147,6 +148,19 @@ def reject_all_values_if_none(source_list)
       end
     end
 
+    def normalize_uri_paths(source_list)
+      source_list.map do |source|
+        # Normalize domains ending in a single / as without omitting the slash accomplisheg the same.
```
vcsjones, line 153: Small nit:

```suggestion
        # Normalize domains ending in a single / as without omitting the slash accomplish the same.
```

lib/secure_headers/headers/content_security_policy.rb
```diff
@@ -147,6 +148,19 @@ def reject_all_values_if_none(source_list)
...
+        # https://www.w3.org/TR/CSP3/#match-paths § 6.6.2.10 Step 2
+        if source.chomp("/").include?("/")
```
vcsjones, line 155: One thing this might not be normalizing is a CSP directive like `https://www.example.com/` since the protocol includes `/`.

The quoted spec is about matching on the `path-part`, not the entire URL that is in the directive.

It would be good to address this, with unit tests as well.

lib/secure_headers/headers/content_security_policy.rb
```diff
@@ -147,6 +148,27 @@ def reject_all_values_if_none(source_list)
...
+    def normalize_uri_paths(source_list)
+      source_list.map do |source|
+        # Normalize domains ending in a single / as without omitting the slash accomplishes the same.
+        # https://www.w3.org/TR/CSP3/#match-paths § 6.6.2.10 Step 2
+        begin
+          uri = URI(source)
+          if uri.path == "/"
+            next source.chomp("/")
+          end
+        rescue URI::InvalidURIError
```
boveus, line 160: @keithamus I know you wrote this code several years ago, but do you remember what edge case we are looking catch here?

It seems like we might be able to get away with something like:

```
source_list.map do |source|
    source.chomp("/")
end
```

I believe `.chomp` will either remove the last char if it exists or return the original string unmodified.

keithamus: I don’t think that’s sufficient as it shouldn’t always remove the trailing `/` - only when the entire path is just `/`. For example “example.com/foo/“ should remain as is, but “example.com/“ should be trimmed so it becomes “example.com”. 

The logic could be described as: if the entire contents of the URL after the domain name are “/“ then that is the equivalent of a URL with no path (so the contents after the domain name are “”).

PR #478, opened 2022-04-01
do not dedupe alternate schema source expresions

keithamus:
## All PRs:

* [x] Has tests
* [x] ~~Documentation updated~~ N/A

## Adding a new header

* Is the header supported by any user agent? If so, which? N/A
* What does it do? N/A
* What are the valid values for the header? N/A
* Where does the specification live? N/A

## Adding a new CSP directive

* Is the directive supported by any user agent? If so, which? N/A
* What does it do? N/A
* What are the valid values for the directive? N/A

This correct a bug in the `dedup_source_list` function, which causes sources with different protocols to be removed, which breaks possible CSP as headers as the spec only allows for _current protocol_ for schemaless source expressions.

Consider the following `connect-src: *.example.com wss://*.example.com`. This effectively allows for http and websocket connections. The current version of `secure_headers` will minify this to be `connect-src: *.example.com` which only allows for http connections, _disallowing websocket connections_.

The fix here is to check that the schemes match for wildcard sources, using `URI(source).scheme`. For a scheme-less URI like `*.example.com`, the URI will come back with a `scheme` of `nil`, which is fine. There is a step before `dedup_source_list` which strips known schemes like `http`/`https`.

### DIFF
--- a/lib/secure_headers/headers/content_security_policy.rb
+++ b/lib/secure_headers/headers/content_security_policy.rb
@@ -155,9 +155,10 @@ module SecureHeaders
       wild_sources = sources.select { |source| source =~ STAR_REGEXP }
 
       if wild_sources.any?
+        schemes = sources.map { |source| [source, URI(source).scheme] }.to_h
         sources.reject do |source|
           !wild_sources.include?(source) &&
-            wild_sources.any? { |pattern| URI(pattern).scheme == URI(source).scheme && File.fnmatch(pattern, source) }
+            wild_sources.any? { |pattern| schemes[pattern] == schemes[source] && File.fnmatch(pattern, source) }
         end
       else
         sources
--- a/lib/secure_headers/headers/content_security_policy.rb
+++ b/lib/secure_headers/headers/content_security_policy.rb
@@ -157,7 +157,7 @@ module SecureHeaders
       if wild_sources.any?
         sources.reject do |source|
           !wild_sources.include?(source) &&
-            wild_sources.any? { |pattern| File.fnmatch(pattern, source) }
+            wild_sources.any? { |pattern| URI(pattern).scheme == URI(source).scheme && File.fnmatch(pattern, source) }
         end
       else
         sources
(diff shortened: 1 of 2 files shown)


lib/secure_headers/headers/content_security_policy.rb
```diff
@@ -157,7 +157,7 @@ def dedup_source_list(sources)
       if wild_sources.any?
         sources.reject do |source|
           !wild_sources.include?(source) &&
-            wild_sources.any? { |pattern| File.fnmatch(pattern, source) }
+            wild_sources.any? { |pattern| URI(pattern).scheme == URI(source).scheme && File.fnmatch(pattern, source) }
```
JackMc, line 160: For large fields, this could result in the source URL being parsed a lot. Would it make sense to precompute the source scheme in the prior loop?

JackMc (changes requested):
Hey! Thanks for the contribution. I'm unable to reproduce this behaviour. I set up a standard Rails 7 app with a scaffolded resource named `posts` and added the following SecureHeaders initializer:

```ruby
  config.csp = {
    default_src: %w(* 'unsafe-inline' blob:),
    script_src: %w('self' blob: 'unsafe-inline' data:),
    connect_src: %w('self' *.example.com wss://*.example.com),
  }
end
```

And the following inline script in the resources #index page:

```html
<script>
  fetch('https://example.com');
</script>

<script>
  webSocket = new WebSocket('wss://example.com');
</script>
```

The CSP returned from Secure Headers appears to be:
```http
Content-Security-Policy: default-src * 'unsafe-inline' blob:; block-all-mixed-content; connect-src 'self' *.example.com wss://*.example.com; script-src 'self' blob: 'unsafe-inline' data:
```

Copying the test from `spec/lib/secure_headers/headers/content_security_policy_spec.rb` to `master` passes. Could you provide a failing test case? Happy to hop on a call and discuss if that's easier!

keithamus:
Thanks for discussing over Slack @JackMc! I've addressed both the badly written test and the precomputation of schemes.

JackMc (approved):

JackMc:
I don't see any problem with it!

srt32:
I don't have merge access. Could you click the button?

PR #479, opened 2022-04-01
Update default X-XSS-Protection value to 0

rzhade3:
This PR updates the default value of the `X-XSS-Protection` header to 0. There's further discussion here about the reasons for this: https://github.com/github/secure_headers/issues/439. 

## All PRs:

* [x] Has tests
* [x] Documentation updated

Closes https://github.com/github/secure_headers/issues/439


JackMc (commented):
I agree that this would be a good change and should be merged as part of the next major version. I started a tracking issue to collect all the things that would go into that version: https://github.com/github/secure_headers/issues/480.

richter-alex:
Howdy @JackMc! Any updates on when we might be able to expect a new major release being cut with this in it?

boveus:
We might also want to update https://github.com/github/secure_headers/blob/main/README.md?plain=1#L53 as part of this PR. This isn't the default config per se, but some folks interpret  it as such when they set the gem up.

boveus (approved):

vcsjones (approved):

PR #484, opened 2022-06-01
Handle different schemes in dedup logic

srt32:
closes https://github.com/github/secure_headers/issues/317

## All PRs:

* [X] Has tests
* [ ] Documentation updated - should anything be updated?

## Adding a new header

* What does it do?

## Adding a new CSP directive

* What does it do?

## Some questions

* Does this change follow the spec?
* Is `URI.parse` okay? Could it raise? Is there a more performant way to do this?


srt32:
https://github.com/github/secure_headers/pull/478 beat me to it!

PR #486, opened 2022-06-08
Add trusted-types and require-trusted-types-for CSP Directive

KyFaSt:
## All PRs:

* [x] Has tests
* [ ] Documentation updated

## Adding a new CSP directive

This PR adds support for the [TrustedTypes](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Security-Policy/trusted-types) CSP directives, [`require-trusted-types-for`](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Security-Policy/require-trusted-types-for) and 'trusted-types'.

TrustedTypes are available in [Chrome 83, Edge 83](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Security-Policy/trusted-types#browser_compatibility) and a [polyfill](https://github.com/w3c/webappsec-trusted-types#polyfill) exists for browsers that don't currently support.
* What does it do?
`require-trusted-types-for` disallows using strings with DOM XSS [injection sink functions](https://w3c.github.io/webappsec-trusted-types/dist/spec/#injection-sinks) (functions like Document.write, Element.InnerHtml), and requires matching types [created by Trusted Type policies](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Security-Policy/require-trusted-types-for#examples). 

* What are the valid values for the directive
`require-trusted-types-for`  only allows `script`. 
`trusted-types` allows a list of strings, including one `'allow-duplicates'` which allows multiples of the same named policy to be created.

### DIFF
--- a/spec/lib/secure_headers/headers/content_security_policy_spec.rb
+++ b/spec/lib/secure_headers/headers/content_security_policy_spec.rb
@@ -151,10 +151,6 @@ module SecureHeaders
         expect(csp.value).to eq("default-src 'self'; require-trusted-types-for script")
       end
 
-      it "does not support style for require-trusted-types-for directive" do
-        expect { ContentSecurityPolicy.new({require_trusted_types_for: %(script style)}) }.to raise_error(ContentSecurityPolicyConfigError)
-      end
-
       it "includes prefetch-src" do
         csp = ContentSecurityPolicy.new(default_src: %w('self'), prefetch_src: %w(foo.com))
         expect(csp.value).to eq("default-src 'self'; prefetch-src foo.com")
--- a/spec/lib/secure_headers/headers/content_security_policy_spec.rb
+++ b/spec/lib/secure_headers/headers/content_security_policy_spec.rb
@@ -146,6 +146,15 @@ module SecureHeaders
         expect(csp.value).to eq("default-src 'self'; require-sri-for script style")
       end
 
+      it "supports require-trusted-types-for directive" do
+        csp = ContentSecurityPolicy.new({require_trusted_types_for: %(script)})
+        expect(csp.value).to eq("require-trusted-types-for script")
+      end
+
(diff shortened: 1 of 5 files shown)


spec/lib/secure_headers/headers/content_security_policy_spec.rb
```diff
@@ -141,6 +141,11 @@ module SecureHeaders
         expect(csp.value).to eq("default-src 'self'; require-sri-for script style")
       end
 
+      it "allows style as a require-trusted-types-src" do
```
lgarron, line 144: Nit: I don't think `require-trusted-types-src` is defined?
```suggestion
      it "allows style as a require-trusted-types-for source" do
```

spec/lib/secure_headers/headers/content_security_policy_spec.rb
```diff
@@ -180,6 +185,16 @@ module SecureHeaders
         csp = ContentSecurityPolicy.new({style_src: %w('self'), style_src_attr: %w('self')})
         expect(csp.value).to eq("style-src 'self'; style-src-attr 'self'")
       end
+
+      it "supports trusted-types directive" do
+        csp = ContentSecurityPolicy.new({trusted_types: %w(blahblahpolicy)})
+        expect(csp.value).to eq("trusted-types blahblahpolicy")
+      end
+
+      it "allows duplicate policy names in trusted-types directive" do
```
lgarron, line 194: Would it make sense to test `'none'` explicitly as well?

lgarron (approved):
Looks good to me!

A few small optional nits.

spec/lib/secure_headers/headers/content_security_policy_spec.rb
KyFaSt: oof, you're right. thanks!

PR #490, opened 2022-06-29
fix source dedup breaking with port wildcards

machisuji:
Fixes the bug introduced with release v6.3.4 in #478 that causes errors like the following when using port wildcards such as `ws://localhost:*` in one's sources:

```
2022-06-29 08:56:52 +0100 Rack app ("GET /" - (::1)): #<URI::InvalidURIError: bad URI(is not URI?): "ws://localhost:*">
2022-06-29 08:56:52 +0100 Rack app ("GET /favicon.ico" - (::1)): #<URI::InvalidURIError: bad URI(is not URI?): "ws://localhost:*">
```

### DIFF
--- a/lib/secure_headers/headers/content_security_policy.rb
+++ b/lib/secure_headers/headers/content_security_policy.rb
@@ -154,10 +154,10 @@ module SecureHeaders
     # e.g. *.github.com asdf.github.com becomes *.github.com
     def dedup_source_list(sources)
       sources = sources.uniq
-      wild_sources = sources.select { |source| source =~ STAR_REGEXP }
+      wild_sources = sources.select { |source| source =~ DOMAIN_WILDCARD_REGEX || source =~ PORT_WILDCARD_REGEX }
 
       if wild_sources.any?
-        schemes = sources.map { |source| [source, URI(source).scheme] }.to_h
+        schemes = sources.map { |source| [source, source_scheme(source)] }.to_h
         sources.reject do |source|
           !wild_sources.include?(source) &&
             wild_sources.any? { |pattern| schemes[pattern] == schemes[source] && File.fnmatch(pattern, source) }
@@ -211,6 +211,14 @@ module SecureHeaders
 
     def symbol_to_hyphen_case(sym)
       sym.to_s.tr("_", "-")
+    end
+
+    def source_scheme(source)
+      uri = URI(source.sub(PORT_WILDCARD_REGEX, ""))
+      # If host is nil the given source doesn't contain a scheme
+      # e.g. for `example.org:443` it would return `example.org` as the scheme
+      # which is of course incorrect
+      uri.scheme if uri.host
     end
   end
 end
--- a/lib/secure_headers/headers/policy_management.rb
+++ b/lib/secure_headers/headers/policy_management.rb
@@ -152,7 +152,8 @@ module SecureHeaders
 
     FETCH_SOURCES = ALL_DIRECTIVES - NON_FETCH_SOURCES - NON_SOURCE_LIST_SOURCES
 
-    STAR_REGEXP = Regexp.new(Regexp.escape(STAR))
+    DOMAIN_WILDCARD_REGEX = /(?<=\A|[^:])\*/
+    PORT_WILDCARD_REGEX = /:\*/
     HTTP_SCHEME_REGEX = %r{\Ahttps?://}
 
     WILDCARD_SOURCES = [
--- a/spec/lib/secure_headers/headers/content_security_policy_spec.rb
+++ b/spec/lib/secure_headers/headers/content_security_policy_spec.rb
@@ -56,6 +56,16 @@ module SecureHeaders
         expect(csp.value).to eq("default-src *.example.org")
       end
 
+      it "minifies overlapping port wildcards" do
+        csp = ContentSecurityPolicy.new(default_src: %w(example.org example.org:* example.org:443 https://example.org:80))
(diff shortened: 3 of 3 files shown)


vcsjones:
Thanks for the pull request! We'll be looking at this shortly.

lib/secure_headers/headers/content_security_policy.rb
```diff
@@ -212,5 +212,9 @@ def strip_source_schemes(source_list)
     def symbol_to_hyphen_case(sym)
       sym.to_s.tr("_", "-")
     end
+
+    def source_scheme(source)
+      URI(source.sub(PORT_WILDCARD_REGEX, '')).scheme
```
lgarron, line 217: Unfortunately, this calculates the source scheme of `example.org:443` as `example.org`

lib/secure_headers/headers/content_security_policy.rb
```diff
@@ -154,10 +154,10 @@ def reject_all_values_if_none(source_list)
     # e.g. *.github.com asdf.github.com becomes *.github.com
     def dedup_source_list(sources)
       sources = sources.uniq
-      wild_sources = sources.select { |source| source =~ STAR_REGEXP }
+      wild_sources = sources.select { |source| source =~ DOMAIN_WILDCARD_REGEX }
```
lgarron, line 157: Could you explain the reasoning behind this?

As far as I Can tell, it should be fine to remove `example.org:443` if `example.org:*` is present.

lgarron (commented):
Thanks for pointing out the issue, and also sending PR!

It would certainly be valuable not to run into an error with wildcard ports, but I have a few questions about this particular fix.

lib/secure_headers/headers/content_security_policy.rb
lgarron: This might work well enough for us.
```suggestion
      source.split("://")[-2] || source.split(":")[-2]
```

lgarron:
@machisuji Would you be able to revise this PR?

machisuji:
> @machisuji Would you be able to revise this PR?

Oh boy, sorry! I completely forgot about that. I will do that first thing tomorrow morning! Thanks for your patience.

lib/secure_headers/headers/content_security_policy.rb
machisuji: I feel safer still using URI parsing. I've amended `source_scheme` so that it considers this special case where the actual scheme is missing.

machisuji:
@lgarron I've pushed some changes which should address your remarks.

lgarron (approved):
Thanks, and apologies for the delay! We had some internal complications around testing `secure_headers` upgrades.

I think it would be nicer to fully parse URL expressions, but I believe this workaround should be safe enough to avoid the current bug. It may fail to deduplicate some patterns, but this is only a matter of optimization (not correctness) and we can address it in the future.

Thanks for the contribution!

machisuji:
Just pushed a tiny correction to use [double quotes](https://github.com/github/secure_headers/pull/490/files#diff-6f8c2fbede52ee1e1081d604950f5b34dd81518a53a477b64d9a1fdf00ebe5fbR217) where single quotes were used before.

lib/secure_headers/headers/content_security_policy.rb
```diff
@@ -212,5 +212,13 @@ def strip_source_schemes(source_list)
     def symbol_to_hyphen_case(sym)
       sym.to_s.tr("_", "-")
     end
+
+    def source_scheme(source)
+      uri = URI(source.sub(PORT_WILDCARD_REGEX, ""))
+      # If host is nil the given source doesn't contain a scheme
+      # e.g. for `example.org:443` it would return `example.org` as the scheme
+      # which is of course incorrect
+      uri.scheme if uri.host
```
lgarron, line 221: It turns out that this also fails for URLs like `https://*.example.org:*`

What do you think of hardcoding just HTTP and HTTPS to avoid further wildcard issues?
```suggestion
      return "https" if source.start_with?("https://")
      return "http" if source.start_with?("http://")
      return nil
```

machisuji: Yeah I agree. I'll add a test case for that today and change it to the regex you suggested.

machisuji: Actually having looked into it again I don't quite get what the issue is. Parsing `https://*.example.org:*` would indeed fail but this is why in line 217 `uri = URI(source.sub(PORT_WILDCARD_REGEX, ""))` it removes the port wildcard.

```
irb(main):003:0> PORT_WILDCARD_REGEX = /:\*/
=> /:\*/
irb(main):004:0> source = "https://*.example.org:*"
=> "https://*.example.org:*"
irb(main):005:0> URI(source.sub(PORT_WILDCARD_REGEX, ""))
=> #<URI::HTTPS https://*.example.org>
irb(main):006:0> _.scheme
=> "https"
```

lgarron: FYI, I've tried to give this whole code a refactor at https://github.com/github/secure_headers/pull/497

lgarron:
We've removed the feature altogether in https://github.com/github/secure_headers/pull/499

Thanks for the PR; I'm sorry it didn't work out.

machisuji:
@lgarron Thanks for the update. I think it's the right call. If people want to save precious bytes, they can make sure not to pass in duplicate sources themselves, really. I'm happy either way because removing the feature removes the bug that was hindering us.

PR #497, opened 2022-08-17
Semantically parse source expressions.

lgarron:
Recently, we've had a spate of fixes for parsing directives and source-expressions, stemming from the fact that the code doesn't understand the format of valid expressions, and makes local assumptions about what they look like — in particular, assuming a resemblance to URLs during deduplication.

https://github.com/github/secure_headers/pull/490
https://github.com/github/secure_headers/pull/478

This PR is an attempt to 'bite the bullet" and parse source expressions so we can semantically deduplicate matching URLs.

## All PRs:

* [x] Has tests
* [ ] ~Documentation updated~ (N/A)

### DIFF
--- a/lib/secure_headers/headers/content_security_policy/source_expression.rb
+++ b/lib/secure_headers/headers/content_security_policy/source_expression.rb
@@ -1,13 +1,15 @@
 # frozen_string_literal: true
 
 require_relative "host_source_expression"
-require_relative "host_source_expression"
+require_relative "quoted_source_expression"
+require_relative "scheme_source_expression"
 
 module SecureHeaders
   class ContentSecurityPolicy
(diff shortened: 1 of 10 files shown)


lib/secure_headers/headers/content_security_policy/source_expression.rb
```diff
@@ -0,0 +1,33 @@
+# frozen_string_literal: true
+
+module SecureHeaders
+  class ContentSecurityPolicy
+    class SourceExpression
```
lgarron, line 5: I wonder if we should just call this a "expression" even something more general, rather than a "source expression". Rather than making this sound like it's specifically for a certain subset of possible values, it would include report paths, as well as unquoted directives like `script-src`.

PR #498, opened 2022-08-17
Semantically parse and deduplicate source expressions

lgarron:
Recently, we've had a spate of fixes for parsing directives and source expressions, stemming from the fact that the code doesn't understand the format of valid expressions, and makes local assumptions about what they look like — in particular, assuming a resemblance to URLs during deduplication, when handling a lot of possible values that are not URLs.

https://github.com/github/secure_headers/pull/490
https://github.com/github/secure_headers/pull/478

This PR is an attempt to 'bite the bullet" and parse source expressions so we can semantically deduplicate matching URLs. In the future, we could use this to add more validation.

## All PRs:

* [x] Has tests
* [ ] ~Documentation updated~ (N/A)

### DIFF
--- a/lib/secure_headers/headers/content_security_policy.rb
+++ b/lib/secure_headers/headers/content_security_policy.rb
@@ -109,14 +109,25 @@ module SecureHeaders
     def build_source_list_directive(directive)
       source_list = @config.directive_value(directive)
       if source_list != OPT_OUT && source_list && source_list.any?
-        minified_source_list = minify_source_list(directive, source_list).join(" ")
-
-        if minified_source_list =~ /(\n|;)/
-          Kernel.warn("#{directive} contains a #{$1} in #{minified_source_list.inspect} which will raise an error in future versions. It has been replaced with a blank space.")
+        cleaned_source_list = []
+        semicolon_warned_yet = false
+        source_list.map do |entry|
+          if entry =~ /(\n|;)/
+            if !semicolon_warned_yet
+              Kernel.warn("#{directive} contains a #{$1} in #{source_list.join(" ").inspect} which will raise an error in future versions. It has been replaced with a blank space.")
+              semicolon_warned_yet = true
+            end
+            split_entry = entry.split(/\n|;/)
+            cleaned_source_list.append(split_entry)
+          else
+            cleaned_source_list.append(entry)
+          end
         end
 
-        escaped_source_list = minified_source_list.gsub(/[\n;]/, " ")
-        [symbol_to_hyphen_case(directive), escaped_source_list].join(" ").strip
+        puts cleaned_source_list
+
+        minified_source_list = minify_source_list(directive, cleaned_source_list).join(" ")
+        [symbol_to_hyphen_case(directive), minified_source_list].join(" ").strip
       end
     end
 
--- a/lib/secure_headers/headers/content_security_policy.rb
+++ b/lib/secure_headers/headers/content_security_policy.rb
@@ -156,10 +156,10 @@ module SecureHeaders
     # e.g. *.github.com asdf.github.com becomes *.github.com
     def dedup_source_list(sources)
       sources = sources.uniq
-      wild_sources = sources.select { |source| source =~ STAR_REGEXP }
+      wild_sources = sources.select { |source| source =~ DOMAIN_WILDCARD_REGEX || source =~ PORT_WILDCARD_REGEX }
 
       if wild_sources.any?
-        schemes = sources.map { |source| [source, URI(source).scheme] }.to_h
+        schemes = sources.map { |source| [source, source_scheme(source)] }.to_h
         sources.reject do |source|
           !wild_sources.include?(source) &&
             wild_sources.any? { |pattern| schemes[pattern] == schemes[source] && File.fnmatch(pattern, source) }
@@ -213,6 +213,14 @@ module SecureHeaders
 
     def symbol_to_hyphen_case(sym)
       sym.to_s.tr("_", "-")
+    end
+
+    def source_scheme(source)
+      uri = URI(source.sub(PORT_WILDCARD_REGEX, ""))
+      # If host is nil the given source doesn't contain a scheme
+      # e.g. for `example.org:443` it would return `example.org` as the scheme
+      # which is of course incorrect
+      uri.scheme if uri.host
     end
   end
 end
--- a/lib/secure_headers/headers/content_security_policy.rb
+++ b/lib/secure_headers/headers/content_security_policy.rb
@@ -162,9 +162,10 @@ module SecureHeaders
       wildcard_host_source_expressions = host_source_expressions.select { |source| source.has_wildcard? }
       
       if wildcard_host_source_expressions.any?
-        host_source_expressions.reject do |source|
+        filtered = host_source_expressions.reject do |source|
             wildcard_host_source_expressions.any? { |wilcard_source| wildcard_source.matches_same_or_superset?(source) }
         end
+        filtered.map { |source| source.to_s }
       else
         sources
       end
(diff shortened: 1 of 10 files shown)


spec/lib/secure_headers/headers/content_security_policy_spec.rb
```diff
@@ -29,13 +29,13 @@ module SecureHeaders
       end
 
       it "deprecates and escapes semicolons in directive source lists" do
-        expect(Kernel).to receive(:warn).with(%(frame_ancestors contains a ; in "google.com;script-src *;.;" which will raise an error in future versions. It has been replaced with a blank space.))
-        expect(ContentSecurityPolicy.new(frame_ancestors: %w(https://google.com;script-src https://*;.;)).value).to eq("frame-ancestors google.com script-src * .")
+        expect(Kernel).to receive(:warn).with(%(frame_ancestors contains a ; in "https://google.com;script-src https://localhost;example.com;" which will raise an error in future versions. It has been replaced with a blank space.))
+        expect(ContentSecurityPolicy.new(frame_ancestors: %w(https://google.com;script-src https://localhost;example.com;)).value).to eq("frame-ancestors google.com script-src localhost example.com")
```
lgarron, line 33: Based on the CSP spec/MDN docs, it looks like `.` is not a valid expression? https://www.w3.org/TR/CSP3/#framework-directive-source-list

This test seems to come from https://github.com/github/secure_headers/issues/418

lib/secure_headers/headers/content_security_policy/source_expression.rb
```diff
@@ -0,0 +1,33 @@
+# frozen_string_literal: true
+
+module SecureHeaders
+  class ContentSecurityPolicy
+    class SourceExpression
```
lgarron, line 5: I wonder if we should just call this a "expression" even something more general, rather than a "source expression".

My goal was to support just "source lists", as defined in the spec and implied in `content_security_policy.rb`. But for now we still need "source expression" parsing to allow report paths, as well as unquoted directives like `script-src`.

JackMc:
@machisuji if you have time, your review would also be appreciated here having worked in this area very recently 🙇

JackMc:
Meta-concern: is this change a breaking one? I don’t think we have an easy way to differentiate between the quirks of this implementation and the other one. However technically those quirks weren’t documented, [they likely are now relied upon](https://www.hyrumslaw.com).

lib/secure_headers/headers/content_security_policy.rb
```diff
@@ -108,15 +109,32 @@ def build_media_type_list_directive(directive)
...
+    # - where any expression containing a deprecated character (semicolon or newline) has been split by that
+    # character, and
+    # - any empty string expression is removed.
+    def clean_malformatted_sources(directive, source_list)
+      cleaned_source_list = []
+      semicolon_warned_yet = false
+      source_list.map do |expression|
+        if expression =~ /(\n|;)/
+          if !semicolon_warned_yet
+            Kernel.warn("#{directive} contains a #{$1} in #{source_list.join(" ").inspect} which will raise an error in future versions. It has been replaced with a blank space.")
```
JackMc, line 128: Likely the `$1` in the warning here will be very confusing if it’s an actual new line. Could we do `$1.inspect`? I think that will show `”\n”`

lib/secure_headers/headers/content_security_policy.rb
```diff
@@ -108,15 +109,32 @@ def build_media_type_list_directive(directive)
...
-        if minified_source_list =~ /(\n|;)/
-          Kernel.warn("#{directive} contains a #{$1} in #{minified_source_list.inspect} which will raise an error in future versions. It has been replaced with a blank space.")
+    # Private: Calculates a modified version version of source_list:
+    # - where any expression containing a deprecated character (semicolon or newline) has been split by that
+    # character, and
+    # - any empty string expression is removed.
+    def clean_malformatted_sources(directive, source_list)
+      cleaned_source_list = []
+      semicolon_warned_yet = false
+      source_list.map do |expression|
```
JackMc, line 125: This isn’t a map operation since it doesn’t have a meaningful array result. However we could do this using a [reduce](https://apidock.com/ruby/Enumerable/reduce) pattern, what do you think? `reduce` is a weird name for the operation in this context. Ruby also offers `inject` which is the same thing just aliased.

```ruby
cleaned_source_list = source_list.inject([]) do |arr, source|
  # … add elements to `arr` as needed.
end
```

If we like the code as-is, we could just change this to `each`.

lib/secure_headers/headers/content_security_policy.rb
```diff
@@ -108,15 +109,32 @@ def build_media_type_list_directive(directive)
...
+          split_expression = expression.split(/\n|;/)
+          cleaned_source_list.concat(split_expression)
+        else
+          cleaned_source_list.append(expression)
         end
-
-        escaped_source_list = minified_source_list.gsub(/[\n;]/, " ")
-        [symbol_to_hyphen_case(directive), escaped_source_list].join(" ").strip
       end
+      cleaned_source_list.select { |value| value != "" }
```
JackMc, line 137: I can’t remember if this library only works in Rails, but if it does you can use [`compact_blank`](https://edgeapi.rubyonrails.org/classes/Enumerable.html#method-i-compact_blank) here.

```suggestion
      cleaned_source_list.compact_blank
```

Otherwise, I would do something like:

```suggestion
      cleaned_source_list.reject(&:blank?)
```

Additionally, with the suggestion above returning the array, you could call this on the return value and remove the need for `cleaned_source_list`.

lib/secure_headers/headers/content_security_policy.rb
```diff
@@ -156,17 +174,14 @@ def reject_all_values_if_none(source_list)
     # e.g. *.github.com asdf.github.com becomes *.github.com
     def dedup_source_list(sources)
       sources = sources.uniq
-      wild_sources = sources.select { |source| source =~ STAR_REGEXP }
+      host_source_expressions = sources.map { |source| parse_source_expression(source) }
+      # TODO: Split by source expression type.
```
JackMc, line 178: Is this TODO before the PR merges? If not we should likely document it in a ticket and remove this comment.

lib/secure_headers/headers/content_security_policy.rb
```diff
@@ -156,17 +174,14 @@ def reject_all_values_if_none(source_list)
     # e.g. *.github.com asdf.github.com becomes *.github.com
     def dedup_source_list(sources)
       sources = sources.uniq
-      wild_sources = sources.select { |source| source =~ STAR_REGEXP }
+      host_source_expressions = sources.map { |source| parse_source_expression(source) }
```
JackMc, line 177: If you want to not copy the array several times, you can use the in-place variants of these methods. Generally they end with an `!`.
```suggestion
      sources.map! { |source| parse_source_expression(source) }
```

JackMc: Same for `uniq`/`uniq!` and `select`/`select!`

lib/secure_headers/headers/content_security_policy.rb
```diff
@@ -214,5 +229,9 @@ def strip_source_schemes(source_list)
     def symbol_to_hyphen_case(sym)
       sym.to_s.tr("_", "-")
     end
+
+    def source_scheme(source)
+      source.match(/^([A-Za-z0-9\-\+.]+):\/\//)&.values_at(1)
```
JackMc, line 234: Is doing this using a regex the most reliable way? I imagine we had a good reason for not using a URI parser.

JackMc (commented):
Didn’t make it through the entire PR, but some initial feedback. Will attempt to remember to come back and do more, but please ping me if I forget.

lib/secure_headers/headers/content_security_policy.rb
KyFaSt, line 128: This may be out of scope for this PR but this kernel warning message and cleaning of `;` has been present since [Jan 2020](https://github.com/github/secure_headers/commit/3c4b86edd6745275da22d92290872da202d73e64). Maybe the version in which this change is released is the version that raises an error? If this method raised an error, it could reduce the logic we need to verify in test.

lgarron:
Abandoned in favor of https://github.com/github/secure_headers/pull/499

PR #504, opened 2022-12-26
Adds Ruby 3.2 to the CI matrix

petergoldstein:
## All PRs:

* [X] Has tests
* [X] Documentation updated

Also updated the checkout action version.

To get the tests to pass, I needed to address a couple of minor lints.


lgarron:
Looks like this was added in: https://github.com/ruby/setup-ruby/pull/424

Thanks!

PR #520, opened 2024-08-08
Update build.yml

boveus:
## All PRs:

* [x] Has tests N/A
* [x] Documentation updated N/A

## Adding a new header

* What does it do?

## Adding a new CSP directive

* What does it do?

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -17,7 +17,7 @@ jobs:
           ruby-version: ${{ matrix.ruby }}
       - name: Install Bundler
         run: |
-          if [[ ${{ matrix.ruby }} == '2.6' or  ${{ matrix.ruby }} == '2.7' ]]; then
+          if [[ ${{ matrix.ruby }} == '2.6' || ${{ matrix.ruby }} == '2.7' ]]; then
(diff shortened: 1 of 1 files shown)


.github/workflows/build.yml
```diff
@@ -10,15 +10,20 @@ jobs:
...
-        gem install bundler
-        bundle install --jobs 4 --retry 3 --without guard
-        bundle exec rspec spec
-        bundle exec rubocop
+      - uses: actions/checkout@v2
+      - name: Set up Ruby
+        uses: ruby/setup-ruby@v1
+        with:
+          ruby-version: ${{ matrix.ruby }}
+      - name: Install Bundler
```
rzhade3, line 18: Bundler is automatically installed with the `ruby/setup-ruby` action, including with this conditional.

rzhade3 (changes requested):

rzhade3 (approved):
It appears as if we're running into some Coveralls errors in CI:

Screenshot 2024-08-08 at 7 50 58 PM

We should either fix this error, or remove the dependency on Coveralls. However, 👍 on this PR to fix builds in production.

PR #529, opened 2024-10-18
Add report-to CSP directive

loremotta33:
I added the `secure_headers` gem to my project as we needed to implement CSP. When configuring the policy, I was going to add the `report-to` directive as it seems that `report-uri` has been deprecated. I noticed that the gem does not have support for this configuration so decided to add it.

## All PRs:

* [x] Has tests
* [x] Documentation updated

## Adding a new CSP directive

Is the directive supported by any user agent? If so, which?

[Yes](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Security-Policy/report-to#browser_compatibility), it's supported by all main browsers except Firefox.

What does it do?

It's used to indicate the name of the endpoint that the browser should use for reporting CSP violations. Intends to replace deprecated `report-uri` directive. Browsers should fallback tho `report-uri` if `report-to` not supported, so it's recommended to set both directives.

An array of endpoints is supported. But if more than one endpoint is provided, browser will default to use the first one. I added it to `report-uri` to preserve the scheme, as as far as I checked, there is no restriction on the naming of the endpoint.

This directive is directly tied to the [Reporting-Endpoints](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Reporting-Endpoints) HTTP header, at the moment on my project I'm manually adding the `Reporting-Endpoints` header through the Rails `default_headers` configuration, I do not know if managing this header through the gem is something on the scope of the gem.

[Here](https://w3c.github.io/webappsec-csp/#directive-report-to) is the specification for `report-to`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -81,12 +81,14 @@ SecureHeaders::Configuration.default do |config|
     style_src_attr: %w('unsafe-inline'),
     worker_src: %w('self'),
     upgrade_insecure_requests: true, # see https://www.w3.org/TR/upgrade-insecure-requests/
-    report_uri: %w(https://report-uri.io/example-csp)
+    report_uri: %w(https://report-uri.io/example-csp),
+    report_to: %w(example-csp)
   }
   # This is available only from 3.5.0; use the `report_only: true` setting for 3.4.1 and below.
   config.csp_report_only = config.csp.merge({
     img_src: %w(somewhereelse.com),
-    report_uri: %w(https://report-uri.io/example-csp-report-only)
+    report_uri: %w(https://report-uri.io/example-csp-report-only),
+    report_to: %w(example-csp-report-only)
   })
 end
 ```
--- a/lib/secure_headers/headers/content_security_policy.rb
+++ b/lib/secure_headers/headers/content_security_policy.rb
@@ -46,8 +46,8 @@ module SecureHeaders
     private
 
     # Private: converts the config object into a string representing a policy.
-    # Places default-src at the first directive and report-uri as the last. All
-    # others are presented in alphabetical order.
+    # Places default-src at the first directive, report-uri second to last and report-to as the
+    # last. All others are presented in alphabetical order.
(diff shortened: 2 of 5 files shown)


rzhade3:
Hi @loremotta33, thanks so much for this PR! This looks like a worthy change to upstream, however I just have some confusion around how exactly we should support this directive in the library. 

It looks like to support this directive, we'll also need to support the [`Report-Endpoints`](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Reporting-Endpoints) header? However, this header is not supported in Safari or Opera? I'm slightly worried about causing confusion amongst users of our library due to incompatibility in these browsers; although ultimately given the status of `Report-To` as an accepted CSP directive, this is a good change.

loremotta33:
That's a great catch @rzhade3, I have not noticed that the `report-to` browser compatibility table is different than the `Reporting-Endpoints` one.

I'm not sure if the tables are correct or up to date though, as I just downloaded the latest Opera version (`114.0.5282.115`), and it has used the Endpoint defined on the `Reporting-Endpoints` header to report the CSP violation (screenshot attached).

image

If you think it makes sense to add support to the `Reporting-Endpoints` header through the gem let me know and I can take a look to the code to do so! 

I thought of initially adding support for it on this Pull Request but I was not sure if it was something that would be wanted or not for the gem

rzhade3:
@loremotta33 Thanks for doing that research! Given that it is strictly _needed_ to implement `Reporting-Endpoints` in order for `report-to` to be respected by a browser, let's try and add that to this PR! Can you take a look at whether you're able to implement that?

loremotta33:
@rzhade3 , sorry for the delay on this, I have been pretty busy the last days.

I just pushed a commit to add `Reporting-Endpoints` support, did some testings myself and I think that it's working as expected, but let me know if some extra work is needed for this!

PR #546, opened 2025-02-06
fix: Fix typos

myersg86:
## All PRs:

* [ ] Has tests
* [ ] Documentation updated

## Adding a new header

* What does it do?

## Adding a new CSP directive

* What does it do?


fletchto99 (approved):

fletchto99 (approved):
Thanks for the fixes 🎉

PR #551, opened 2025-03-20
Remove non-lowercase headers in Rails default configuration (fixes #541)

### LINKED ISSUE #541: Lowercase header issue: SecureHeaders::OPT_OUT Fails to Remove Non-Lowercase Headers in Rails Default Config
# Bug
Gem version: 7.1.0
Rails: 6.1
Ruby: 3.3.6

Following this change from this [pull-request](https://github.com/github/secure_headers/pull/533)
It introduce a regression and an unwanted behavior with some headers.

While using `SecureHeaders::OPT_OUT` as a value to override the X-Frame-Options header we're not deleting the header.

It happens because here in [railtie.rb](https://github.com/github/secure_headers/blob/8b1029ca6d9a424b903b5beebcd88cbdadf43f4b/lib/secure_headers/railtie.rb#L27)
we are only looking for the lowercase key values.
The default written `X-Frame-Options` never get match because keys are case sensitive.

This is a potential bigger issue since looking a recent version of rails action_dispatch [here](https://github.com/rails/rails/blob/8-0-stable/actionpack/lib/action_dispatch/railtie.rb) it is still using the non lowercase keys. 

### Expected outcome

1. The header is removed from the headers list in the response when using `SecureHeaders::OPT_OUT`.

### Actual outcome

1. Any Non downcase header with the `SecureHeaders::OPT_OUT` value doesn't get remove.

Suggestions:

```ruby
default_headers = Rails.application.config.action_dispatch.default_headers
unless default_headers.nil?
  default_headers.each_key do |header|
    if conflicting_headers.include?(header.downcase)
      default_headers.delete(header)
    end
  end
end
```

obrie:
This implements the solution proposed in https://github.com/github/secure_headers/issues/541

The full details of the issue can are described in that ticket.  To summarize:

While secure_headers now uses lowercase headers (as required by Rack 3+), the Rails default configuration still defines non-lowercase headers.  As a result, our Railtie will not remove those conflicting headers.

This change ensures that we're accounting for both lowercase and non-lowercase default headers in Rails (for current Rails defaults and future defaults).

## All PRs:

* [ ] Has tests
* [x] Documentation updated

## Adding a new header

* What does it do?

## Adding a new CSP directive

* What does it do?

### DIFF
--- a/lib/secure_headers/railtie.rb
+++ b/lib/secure_headers/railtie.rb
@@ -22,9 +22,12 @@ if defined?(Rails::Railtie)
         ActiveSupport.on_load(:action_controller) do
           include SecureHeaders
 
-          unless Rails.application.config.action_dispatch.default_headers.nil?
-            conflicting_headers.each do |header|
-              Rails.application.config.action_dispatch.default_headers.delete(header)
+          default_headers = Rails.application.config.action_dispatch.default_headers
+          unless default_headers.nil?
+            default_headers.each_key do |header|
+              if conflicting_headers.include?(header.downcase)
+                default_headers.delete(header)
+              end
             end
           end
         end


rei-moo (approved):
Looks good, thank you! 🎉

